Issue #183, opened 2013-05-11
http://wpscan.org/ provides erroneous information for installing on Arch Linux


qwerty007:
You should never run 'pacman -Sy foo'  https://bbs.archlinux.org/viewtopic.php?id=89328

```
pacman -Sy ruby
pacman -Sy libyaml
```

should be changed to

```
pacman -Syu ruby libyaml
```

or

```
pacman -Syu ruby 
pacman -Syu libyaml
```

PR #743, opened 2014-12-19
Remove updating of Arch repositories

Sacro:
It's not recommended as it can break dependencies.


ethicalhack3r:
We changed it that after another user recommended it - https://github.com/wpscanteam/wpscan/issues/183

ethicalhack3r:
Closing as I am not sure which method is best as I'm not an Arch user

PR #762, opened 2015-01-28
Add patch dependency to fedora installation

Pablohn26:
It is needed to install nokogiri

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ From Ubuntu 14.04:
 
 ####Installing on Fedora:
 
-    sudo yum install gcc ruby-devel libxml2 libxml2-devel libxslt libxslt-devel libcurl-devel
+    sudo yum install gcc ruby-devel libxml2 libxml2-devel libxslt libxslt-devel libcurl-devel patch
     git clone https://github.com/wpscanteam/wpscan.git
     cd wpscan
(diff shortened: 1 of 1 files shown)


ethicalhack3r:
How did you install Ruby? Looks like patch is already installed when Ruby 2.2.0 is installed with RVM.

Pablohn26:
From Fedora 20 repo (I guess). 
My ruby version is 

     ruby --version
     ruby 2.0.0p353 (2013-11-22 revision 43784) [x86_64-linux]

and from yum

>$ yum info ruby     
>Loaded plugins: langpacks
>Installed Packages
>Name        : ruby
>Arch        : x86_64
>Version     : 2.0.0.353
>Release     : 16.fc20
>Size        : 63 k
>Repo        : installed
>From repo   : updates
>Summary     : An interpreter of object-oriented scripting language
>URL         : http://ruby-lang.org/
>License     : (Ruby or BSD) and Public Domain
>Description : Ruby is the interpreted scripting language for quick and easy
>            : object-oriented programming.  It has many features to process text
>            : files and to do system management tasks (as in Perl).  It is simple,
>            : straight-forward, and extensible.

ethicalhack3r:
Cool, thanks for the heads up! :)

PR #774, opened 2015-02-21
Available take username for wordpress 3.0 or lower.

berotti3:
if username is "User"
wordpress3.0 or lower:
class="archive author author-User">
wordpress3.1 or higher:
class="archive author author-User author-1">

and,if user logged in,
class="archive author author-User author-1 logged-in">

we dont need author-number,logged-in status.so reversion regular expression.


erwanlr:
Good catch, Thanks !

Issue #757, opened 2015-01-20
Add option to modify "We could not determine a version so all vulnerabilities are printed out"


hitl0r:
If a plugin/theme version cannot be determined, wpscan prints out all vulnerabilites for all versions with the message "We could not determine a version so all vulnerabilities are printed out". 
This is too inaccurate imho.

There should be a scan option to skip this behaviour, like

wpscan.rb --strict

which would not print *any* vulnerabilities if the version of a plugin/theme cannot be determined.

The output then would be:
"We could not determine a version and the option "--strict" was used. No vulnerabilities are printed out"

ethicalhack3r:
What we could do is not print out all vulnerabilities by default, but only do it when verbose is used. 

Thoughts?

ethicalhack3r:
Looks like https://github.com/wpscanteam/wpscan/commit/d84335e251b04b6e2691e0d9f02362d79ee242c3 fixes this

PR #783, opened 2015-03-14
Only show all vulns when verbose #757

ethicalhack3r:
If happy please merge @erwanlr @FireFart


erwanlr:
I'm not.

During the enumeration, requests are not cached, so if you do a first scan w/o the -v, potential vulns will not be shown if the version is unknown, then you are good to redo a scan with the -v option, which will take more time than manual check the version / eliminate vulns

Issue #793, opened 2015-04-02
wpstools does nothing?


phryneas:
Hi,
either I'm reading the documentation wrong or wpstools currently does nothing?

I'm running it like this:

```
# ruby wpstools.rb --clvf /var/www/vhosts/known-vulnerable.site/httpdocs/
_______________________________________________________________
        __          _______   _____                  
        \ \        / /  __ \ / ____|                 
         \ \  /\  / /| |__) | (___   ___  __ _ _ __  
          \ \/  \/ / |  ___/ \___ \ / __|/ _` | '_ \ 
           \  /\  /  | |     ____) | (__| (_| | | | |
            \/  \/   |_|    |_____/ \___|\__,_|_| |_|

        WordPress Security Scanner by the WPScan Team 
                       Version 2.7
          Sponsored by Sucuri - https://sucuri.net
   @_WPScan_, @ethicalhack3r, @erwan_lr, pvdl, @_FireFart_
_______________________________________________________________

[+] Generating local hashes ... done.
[+] Checking for vulnerable files ...
done.
```

Running wpscan on the same site from outside works fine:

```
 ruby wpscan.rb --url www.known-vulnerable.site
_______________________________________________________________
        __          _______   _____                  
        \ \        / /  __ \ / ____|                 
         \ \  /\  / /| |__) | (___   ___  __ _ _ __  
          \ \/  \/ / |  ___/ \___ \ / __|/ _` | '_ \ 
           \  /\  /  | |     ____) | (__| (_| | | | |
            \/  \/   |_|    |_____/ \___|\__,_|_| |_|

        WordPress Security Scanner by the WPScan Team 
                       Version 2.7
          Sponsored by Sucuri - https://sucuri.net
   @_WPScan_, @ethicalhack3r, @erwan_lr, pvdl, @_FireFart_
_______________________________________________________________

[+] URL: http://www.known-vulnerable.site/
[+] Started: Thu Apr  2 11:56:54 2015

[+] robots.txt available under: 'http://www.known-vulnerable.site/robots.txt'
[!] The WordPress 'http://www.known-vulnerable.site/readme.html' file exists exposing a version number
[+] Interesting header: SERVER: Apache
[+] Interesting header: X-POWERED-BY: PleskLin
[+] XML-RPC Interface available under: http://www.known-vulnerable.site/xmlrpc.php
[...]

PR #802, opened 2015-04-10
Remove wpstools #793

ethicalhack3r:
Specs pass. Would be good if someone else could have a quick look to ensure that I have removed everything.


erwanlr:
Seems good :)

PR #821, opened 2015-05-08
Enhancement based on issue 518

mrnfrancesco:
I added new usernames enumerating flag to show display names even if no login name is provided.
New flag can be used as usual with `--enumerate u` option with an extra `d` character as following:

```
ud   usernames from id 1 to 10
ud[10-20] usernames from id 10 to 20 (you must write [] chars)
```

This could be useful in case when no login name is retrieved, but display names are the same (or very similar) of the login counterpart.

In my case:
- without using `d` flag

```
[+] Enumerating usernames ...
[+] We did not enumerate any usernames
```
- using `d` flag

```
[+] Enumerating usernames ...
[+] Identified the following 2 user/s:
    +----+-------+-------+
    | Id | Login | Name  |
    +----+-------+-------+
    | 1  |       | admin |
    | 2  |       | marco |
    +----+-------+-------+
```

Obviously, `admin` and `marco` are the login name too.

### DIFF
--- a/lib/common/collections/wp_items/detectable.rb
+++ b/lib/common/collections/wp_items/detectable.rb
@@ -25,6 +25,17 @@ class WpItems < Array
       }
       results          = passive_detection(wp_target, options)
 
+      if self == WpUsers
+        inexistent_user = WpUser.new(wp_target.uri, id: 9999999999)
+        request = browser.forge_request(inexistent_user.url, request_params)
+        request.on_complete do |response|
+          error_404_title = WpUser::Existable.display_name_from_body(response.body)
+          exist_options[:error_404_title] = error_404_title if error_404_title
+        end
+        hydra.queue(request)
+        hydra.run
+      end
+
       targets.each do |target_item|
         request = browser.forge_request(target_item.url, request_params)
 
--- a/lib/common/collections/wp_items/detectable.rb
+++ b/lib/common/collections/wp_items/detectable.rb
@@ -32,7 +32,7 @@ class WpItems < Array
           progress_bar.progress += 1 if options[:show_progression]
 
           if target_item.exists?(exist_options, response)
-            if !results.include?(target_item)
+            unless results.include?(target_item)
               if !options[:only_vulnerable] || options[:only_vulnerable] && target_item.vulnerable?
                 results << target_item
               end
--- a/lib/common/collections/wp_items/detectable.rb
+++ b/lib/common/collections/wp_items/detectable.rb
@@ -19,9 +19,9 @@ class WpItems < Array
       progress_bar     = progress_bar(targets.size, options)
       queue_count      = 0
       exist_options    = {
-        error_404_hash:  wp_target.error_404_hash,
-        homepage_hash:   wp_target.homepage_hash,
-        exclude_content: options[:exclude_content] ? %r{#{options[:exclude_content]}} : nil
+          error_404_hash:  wp_target.error_404_hash,
+          homepage_hash:   wp_target.homepage_hash,
+          exclude_content: options[:exclude_content] ? %r{#{options[:exclude_content]}} : nil
       }
       results          = passive_detection(wp_target, options)
 
@@ -31,7 +31,7 @@ class WpItems < Array
         request.on_complete do |response|
           progress_bar.progress += 1 if options[:show_progression]
 
-          if target_item.exists?(exist_options, response) or (target_item.class.to_s == 'WpUser' and options[:display_name])
+          if target_item.exists?(exist_options, response) or (target_item.kind_of? WpUser and options[:display_name])
             unless results.include?(target_item)
               if !options[:only_vulnerable] || options[:only_vulnerable] && target_item.vulnerable?
                 results << target_item
@@ -67,9 +67,9 @@ class WpItems < Array
     def progress_bar(targets_size, options)
       if options[:show_progression]
         ProgressBar.create(
-          format: '%t %a <%B> (%c / %C) %P%% %e',
-          title: '  ', # Used to craete a left margin
-          total: targets_size
+            format: '%t %a <%B> (%c / %C) %P%% %e',
+            title: '  ', # Used to craete a left margin
+            total: targets_size
         )
       end
     end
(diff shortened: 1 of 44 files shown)


lib/common/collections/wp_items/detectable.rb
```diff
@@ -31,7 +31,7 @@ def aggressive_detection(wp_target, options = {})
         request.on_complete do |response|
           progress_bar.progress += 1 if options[:show_progression]
 
-          if target_item.exists?(exist_options, response)
+          if target_item.exists?(exist_options, response) or options[:display_name]
```
erwanlr: Bad things will happen due to that, i.e with ```-e vp,ud```, all known vulnerable plugins will be reported as present on the blog

lib/common/collections/wp_items/detectable.rb
```diff
@@ -31,7 +31,7 @@ def aggressive_detection(wp_target, options = {})
         request.on_complete do |response|
           progress_bar.progress += 1 if options[:show_progression]
 
-          if target_item.exists?(exist_options, response)
+          if target_item.exists?(exist_options, response) or (target_item.class.to_s == 'WpUser' and options[:display_name])
```
FireFart: why class.to_s instead of kind_of?

lib/common/models/wp_user/existable.rb
```diff
@@ -76,7 +76,10 @@ def self.display_name_from_body(body)
         # replace UTF chars like  &#187; with dummy character
         title_tag.gsub!(/&#(\d+);/, '|')
 
-        name = title_tag[%r{([^|«»]+) }, 1]
+        name = title_tag[%r{.+» (.+)}, 1]
```
FireFart: the problem with this chars is, they are theme dependant. If you use another theme, the chars may be missing or different.
Maybe we can improve the regex like search for word characters as the name instead of relying on the special chars. Or combine both methods

lib/common/collections/wp_items/detectable.rb
mrnfrancesco: Because this is the first time I use Ruby (and I never studied it before, but I will).
By the way, I'll replace it with a `require` and the `kind_of` method. Is it the right way to do it?

lib/common/models/wp_user/existable.rb
mrnfrancesco: In my opinion, actually the best way is to look at the title in some page (e.g. a `404 Not Found` error page) and look for changes.

For example, in my case, the title of a 404 error page is: `Macroarea di Scienze M.F.N. » Pagina non trovata.`. The title asking for author number 1 is `Macroarea di Scienze M.F.N. » admin`.

Saving the substring which is equal and removing it from the author page title we could get the username with no regex or theme dependent issues.

erwanlr: I like this method better than the current checks of the display name (ours and yours) even if it's still theme dependent, but negate the regexp madness (sometimes the display name appears first, e.g: 'admin « Some Title')

mrnfrancesco:
I implemented a sketch of the logic I explained to identify display names.
It seems to work pretty well in cases.

It has just one problem in this case:
```
error_404_title = "Page Not Found - Website Name"
title_tag = "username, Author at Website Name"
```
My piece of code delete correctly ` Website Name` from `title_tag`, but the resulting name is `username, Author at`.
If there is more than one user this problem is solved by your `get_equal_string_end` function in `lib/common/common_helper.rb`, but still exists if the website have just one user registered.

mrnfrancesco:
I tested my last fixes with many websites using different themes and I always caught the right display names.
What do you think about this approach?

erwanlr:
What about this one guys ? @wpscanteam/owners 

Haven't tested it on target with this behaviour, but it seems ok (I played a bit with a stand alone script to try to reduce the code detecting the common part of the titles, but there isn't, strings have to be checked with the old school iteration over each chars :o).

Is there a plugin that produces this behaviour ?

FireFart:
Why not use it as the default behaviour? The user exists and if we can not determine the login, we should also print it (because the user exists)

lib/common/collections/wp_items/detectable.rb
```diff
@@ -25,13 +25,24 @@ def aggressive_detection(wp_target, options = {})
       }
       results          = passive_detection(wp_target, options)
 
+      if self == WpUsers
+        inexistent_user = WpUser.new(wp_target.uri, id: 9999999999)
+        request = browser.forge_request(inexistent_user.url, request_params)
+        request.on_complete do |response|
```
FireFart: i think there is no need for hydra here. Browser.get should work here too?

erwanlr: :+1

lib/common/collections/wp_users/output.rb
```diff
@@ -28,18 +28,10 @@ def output(options = {})
...
-        unless display_name == 'empty'
-          display_names << display_name
-        end
-      end
-      junk = get_equal_string_end(display_names)
-      unless junk.nil? or junk.empty?
-        self.each do |u|
-          u.display_name ||= ''
-          u.display_name = u.display_name.sub(/#{Regexp.escape(junk)}$/, '')
+        unless u.display_name == 'empty'
```
FireFart: the methods you deleted were the "fallback" when the title could not be extracted. It took all usernames and remove the last chars which were the same. I think we should leave it in as a fallback since the chars you remove are totally theme dependent. Each theme can define another title format (even for each site)

mrnfrancesco: I removed it because I scanned some websites with all usernames ending with the same couple of characters with wrong results as display names.
In this way I checked as many website (with different themes) as I could and I always retrieved all the right display names.
This is because of the `inexistent_user` trick: it is theme dependent, but, basically, it does the same of these deleted lines, but in a different moment (before getting users list instead of after that).

Why I preferred this way? because in the worst case there will be no deletion in the title (no common words) and all the title will be printed, letting the user to understand which is the significant part.

mrnfrancesco:
I rebased my branch to solve merge conflicts (sorry for the long list of commits above).

ethicalhack3r:
Hi, Sorry your PR wasn't merged. Due to how old this one is I'm deleting it.

PR #829, opened 2015-06-01
Fix for malformed location: headers in redirects.

infernusdomoso:
I ran into a website that had a bad location: 302 redirect at the landing page. It began with a '?' - and wpscan assumes that unless the redirect begins with a '/' (in which case the redirect is rewritten to a complete URI) that it is properly formed. 

This URI is apparently checked before you are asked if you want to follow the redirect, as I was attempting to scan a site that redirected to ?key=abcd, and wpscan crashed because it was attempting to parse http://?key=abcd before it asked if I wanted to follow the redirect.

By adding a check for a redirect that begins with '?', and if so, to prepend a '/' to it, solved the issue.

### DIFF
--- a/lib/wpscan/web_site.rb
+++ b/lib/wpscan/web_site.rb
@@ -55,6 +55,9 @@ class WebSite
     redirected_uri = URI.parse(add_trailing_slash(add_http_protocol(url)))
     if response.code == 301 || response.code == 302
       redirection = response.headers_hash['location']
+      if redirection[0] == '?'
+        redirection = "/#{redirection}"
+      end
       if redirection[0] == '/'
         redirection = "#{redirected_uri.scheme}://#{redirected_uri.host}#{redirection}"
       end


ethicalhack3r:
Thanks for the pull request!

Just a note on this, according to RFC 2616 this seems to be non-compliant behaviour. However, I have not tested how browsers behave, I assume, the same way you are suggesting. If this the case then I think we should also implement this.

http://tools.ietf.org/html/rfc2616#section-14.30

infernusdomoso:
Oh I'm sure it's non-compliant, but browsers (at least chrome, I havent tested others) just append the ?whatever=something to the end of the URL, so it works. Wpscan, however, bombs on it and it makes the otherwise-working site impossible to scan.

lib/wpscan/web_site.rb
```diff
@@ -55,6 +55,9 @@ def redirection(url = nil)
     redirected_uri = URI.parse(add_trailing_slash(add_http_protocol(url)))
     if response.code == 301 || response.code == 302
       redirection = response.headers_hash['location']
+      if redirection[0] == '?'
+        redirection = "/#{redirection}"
```
FireFart: the redirection should be the same as on / so you extend the if below to smth like `if / or ?`

erwanlr: You could even take advantage of the URI#merge method to replace those two if:

```
2.2.1 :001 > require 'URI'
 => true 
2.2.1 :002 > uri = URI.parse 'https://g.com/dir/'
 => #<URI::HTTPS https://g.com/dir/> 
2.2.1 :003 > uri.merge('/root')
 => #<URI::HTTPS https://g.com/root> 
2.2.1 :004 > uri.merge('?param=1')
 => #<URI::HTTPS https://g.com/dir/?param=1> 
2.2.1 :005 > uri.merge('//h.org')
 => #<URI::HTTPS https://h.org> 
2.2.1 :006 > uri.merge('http://i.net')
 => #<URI::HTTP http://i.net>
```

Btw, in your case, if the redirected_ui is a subdir (e.g http://g.com/dir/) and the redirect something like ?p=2, the redirection returned will be https://g.com/?p=2 rather than the correct one. Using the URI#merge mentioned above will take care of such case ;)

erwanlr:
Implemented the redirected_uri#merge ;)

PR #830, opened 2015-06-03
Fix issue 815

mrnfrancesco:


mrnfrancesco:
In my opinion, the `Time.parse('2000-01-01')` expedient is better located into `update_required?` because in `last_update` it would give a wrong information.

If I just downloaded wpscan and run `wpscan --version`, I think it's better to see nothing instead of a date which is obviously a fake.

ethicalhack3r:
:+1:

ethicalhack3r:
I'll just wait on a second opinion before merging. Thanks for the pull request! :)

FireFart:
looks good to me

PR #862, opened 2015-09-07
Apiv2

ethicalhack3r:


FireFart:
I would change the text from ` |  Latest version: 2.3.8` to ` |  Latest version available: 2.3.8`

FireFart:
Or even only print it if latest_version > current_version and say something like `update available`

ethicalhack3r:
Version output look better @FireFart ?

ethicalhack3r:
Going to use ```#lesser_or_equal?```

ethicalhack3r:
Implemented new method called ```lesser?```

Issue #775, opened 2015-02-25
Feature request: Further throttling of attempts during dictionary attack


cmavr8:
So, I'm running WPScan on a raspberry pi, over a DSL connection, using option --max-threads 1 to scan a remote web server. Still, the server gets overloaded (load 40) at some point.

The initial theme/plugin detection part part of the scan is completed successfully, and usually the dictionary attack on a couple of users completes, too, but while further users are tested I get continuous errors like this one:
ERROR: Server error, try reducing the number of threads.

And I know it's not the server's firewall's problem, I've taken care of that.

Shouldn't there be a way to limit requests per second? A workaround would be to do it externally (cgroups etc) but I'd prefer an option built-in to WPScan. Does anyone else find this useful or should I look for a local solution?

(Related to https://github.com/wpscanteam/wpscan/issues/127, which does not help though)

ethicalhack3r:
Thanks for the feature request! @erwanlr what do you think, something worth implementing?

Would mean adding a new option ```--throttle``` taking the number of seconds to wait before each request. Would probably be useful to be able to provide floats as well so can specify 30 seconds as ```0.5```.

cmavr8:
Yeap, throttling would be exactly what's needed (for me and in my opinion).

I'd go for milliseconds or seconds (float). I don't think anyone will use a rate of more than a few seconds per request (a dictionary attack will take a loooong time). On the other hand I suspect that I'd avoid server overload at around 10 request per second, so one every 100ms.

Zmetta:
A broad request delay option would be very useful for more than just throttling brute force and enumeration. Being able to tune the frequency of all requests coming from wpscan would be extremely useful.

erwanlr:
Implemented in the latest commit thanks to @Becojo

Please use --threads 1 along with the --throttle option (which expects a value in milliseconds)

Haxel0rd:
Note: 
in 2023, "--threads" was not recognised anymore (wpscan v3.8.22). 
Alternatively you can use "--max-threads" instead.

PR #871, opened 2015-10-05
Add --throttle option

Becojo:
I needed this feature quickly (it was also requested in #775) so it isn't probably the best way to implement this.

This options allow to throttle requests made by WPScan by a specified number of milliseconds. This should be used with --threads/--max-threads set to 1.

Some whitespace was also trimmed in the process.

### DIFF
--- a/lib/common/browser.rb
+++ b/lib/common/browser.rb
@@ -17,7 +17,8 @@ class Browser
     :proxy_auth,
     :request_timeout,
     :connect_timeout,
-    :cookie
+    :cookie,
+    :throttle
   ]
 
   @@instance = nil
@@ -76,6 +77,7 @@ class Browser
     @request_timeout = 60 # 60s
     @connect_timeout = 10 # 10s
     @user_agent = "WPScan v#{WPSCAN_VERSION} (http://wpscan.org)"
+    @throttle = 0
   end
 
   #
@@ -135,7 +137,7 @@ class Browser
         @basic_auth
       )
     end
-    
+
     if vhost
       params = Browser.append_params_header_field(
         params,
@@ -143,7 +145,7 @@ class Browser
         vhost
       )
     end
-    
+
     params.merge!(referer: referer)
     params.merge!(timeout: @request_timeout) if @request_timeout
     params.merge!(connecttimeout: @connect_timeout) if @connect_timeout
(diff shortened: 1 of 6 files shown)


erwanlr:
The sleep trick works with 'standalone' request, however as soon as Hydra is used, it's not working as expected (you can use the --proxy option with tools such as Burp suite to check the request time):

Requests in the queue ran at the same time, then the sleep happens (even with a --max-threads of 1, which is weird)

I also tried to use the Typhoeus.on_complete hook but it was worse as requests were still done at the same time (in parallel), then number-of-rq * sleep(throttle) were performed (for example, requests for the wp-config check took less than a second to do all of the 17, then 17 sleep of 1s were done).

This is even weirder as it seems to work with a standalone script such as
```ruby
require 'typhoeus'

@throttle = 2000 / 1000.0

Typhoeus.on_complete do |response|
  next if response.cached?

  next unless @throttle > 0

  sleep @throttle
end

hydra = Typhoeus::Hydra.new(max_concurrency: 1)

(1..10).each do |index|
  rq = Typhoeus::Request.new('http://wp.lab/', proxy: 'http://192.168.1.10:8181')

  rq.on_complete do |res|
    puts res.effective_url
  end

  hydra.queue(rq)
end

hydra.run
```

How did you checked that requests were correctly delayed ?

Becojo:
To test the implementation, I did an enumeration of plugins with a throttle of 250 ms on a target website and it went fine. I set both the `max-threads` and `threads` to 1 to ensure there would be no parallel requests. I also tried that is was throttling correctly on a local installation of Wordpress. 

In your example, the only way I found to throttle those requests is by calling `hydra.run` after each requests is added to the queue and then sleep. That is why I had to set the threads to 1 for my scan to work. 

I guess that specifying a throttle should imply that there is only one thread.

It looks like I have forget to throttle the `hydra.run` there https://github.com/Becojo/wpscan/blob/throttle/lib/common/models/wp_user/brute_forcable.rb#L58

Becojo:
So this last commit fixed the password bruteforce being done without throttling. 

Here's an example output from the server and WPScan:  https://gist.github.com/Becojo/402ca6654a5566ce0857

erwanlr:
Okkkkk, found the reason why it wasn't working on my side ;D

There are two options to set the same thing (threads), one of them is not considered in the code (--max-threads) and I was using it, hence the inconsistency in results xD (https://github.com/wpscanteam/wpscan/issues/872 opened)

I pushed a throttle branch using your method but within the Typhoeus.on_complete hook rather than in hydra loops etc :), could you test it and tell me if it's ok with you ? (I tested it and it works, using the --threads option)

Becojo:
Seems fine!

erwanlr:
merged the throttle branch :)

Thx :)

PR #889, opened 2015-11-24
Add fallback json library

pdufour:
Useful for environments like jruby where c-exts are disabled.


michaelkruglos:
This should be merged, not only for this reason, but also, because yajl-ruby is broken. You can craft a short json, which will cause it to SIGABORT.
Take a look at this pool request for reference: https://github.com/brianmario/yajl-ruby/pull/156

ethicalhack3r:
Hi, Sorry your PR wasn't merged. Due to how old this one is I'm deleting it.

PR #891, opened 2015-12-03
Add support for local-directory scanning

pdufour:
Add support for scanning local WP installation directory for plugins
This can be helpful if you don't want to hit your server with a ton of requests at once.
It behaves the same way but just checks for directory existance of a plugin / theme, etc.

Use by setting `--wp-local-dir`, i.e.

````
ruby wpscan.rb --url google.com --enumerate vp --wp-local-dir /repositories/google/web
````

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+# editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = spaces
+indent_size = 2
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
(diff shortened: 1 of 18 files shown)


erwanlr:
As the tool is a black box scanner, that's not something we will support.

Using a plugin like https://wordpress.org/plugins/plugin-security-scanner/ will do a better job.
If you still want the cli way, I would suggest to have a look at https://github.com/wp-cli/wp-cli

There is also http://wordstress.org/ which is a mix of the above

ethicalhack3r:
Great effort, thank you, but I agree with Erwan

anantshri:
I agree with @ethicalhack3r and @erwanlr on this. However the use case suggested by @pdufour is also a valid use case.  i wrote https://github.com/anantshri/wpvulndb_commandline for similar use case.

However the code right now is working on apiv1 and still needs to be ported to apiv2 (https://github.com/anantshri/wpvulndb_commandline/issues/4)

PR #907, opened 2016-02-22
Remove license from readme

surfer190:
* Change licence to markdown

---

No one wants to read the licence first, so why is it at the top of the readme?

I have added a link at the bottom of the readme to the licence

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,16 +1,20 @@
-WPScan Public Source License
+#### LICENSE
+
+#### WPScan Public Source License
 
 The WPScan software (henceforth referred to simply as "WPScan") is dual-licensed - Copyright 2011-2016 WPScan Team.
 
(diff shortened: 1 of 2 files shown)


ethicalhack3r:
Personally I like it where it is. There's no excuse if we find someone using WPScan commercially without a commercial license that they did not see it.

surfer190:
I think I get your point of view, you are looking from a business perspective. A user seeing the licence, is more important than quickly getting started and using wpscan.

FireFart:
I also think the user can scroll the few lines over the license.

It's not only from a business perspective. We put the license in place to prevent automated scanner as a service platforms to use our data wihtout proper attribution or giving something back to the community. Sad story but it happened more than one time....

surfer190:
That's a shame

PR #909, opened 2016-02-23
drop ruby 1.9 and 2.0 support, whitespaces

FireFart:
This PR drops support for ruby 1.9 and 2.0 because they're EOL and no one should use it.

I also converted a lot of line endings to unix to be consistent with the rest of the code base. To view only the relevant changes visit https://github.com/wpscanteam/wpscan/pull/909/files?w=1

The "new" url_encode also encodes the `/` character so I had to remove some calls so please test this with some of your sites before merging.


FireFart:
CC @erwanlr @ethicalhack3r

ethicalhack3r:
:+1:

FireFart:
Ok I need to update the PR again:
https://www.ruby-lang.org/en/news/2016/02/24/support-plan-of-ruby-2-0-0-and-2-1/

ethicalhack3r:
Although not "officially" supported by us, we could allow users to use 2.0 for now and deprecate it in the future?

I think deprecating 1.9 is a good idea now, but deprecating 2.0 can be done in a few months.

FireFart:
I think we should only support supported versions of ruby. This way you can also inform the users that they use an outdated version of ruby and need to upgrade.
Security guys using EOL versions should rethink their doing anyway :)

ethicalhack3r:
Good points, I agree.

ethicalhack3r:
OK to merge?

FireFart:
if we don't get any more comments - yeah :)

Issue #704, opened 2014-09-30
give dates and current versions for WP


digininja:
As part of a scan you output the version of WP detected, it would be good if you could also say when that version was released and give the current version with its release date. When I'm writing my report I want to put:

You are running version x.x which was released y/y/y, it is recommended you upgrade to the latest which is currently version p.p.

The "shock" value of pointing out that their current version is 2 years out of date (recent test) works well to get them to update and if you've got the data then it saves me going looking it up.

Same for plugins and their dates would be really good but that would be a lot of work.

pvdl:
Made some improvements on the pluginversion script.
https://github.com/wpscanteam/Tools/blob/master/pluginversion.rb

pvdl:
Improved the script a bit more.

Output example:
```
URL         : http://plugins.svn.wordpress.org/woocommerce/trunk/readme.txt
Plugin      : woocommerce
Version     : 2.2.10
Requires WP : 3.8
Tested up to: 4.1
Last update : 2014-12-16
```

@wpscanteam/owners , what do you think, shall we implement this 'code-snippet' in the WPScan-core?
We can use the verbose mode or create an extra option e.g. ```--dates```

It uses the external wordpress svn repository for getting the dates from the installed plugins.

ethicalhack3r:
I like the optional aspect, but maybe using the WordPress web Plugin DB would be faster and likely to have less issues accessing?

FireFart:
i think having this in wpvulndb and pushing it to the json makes more sense to reduce the number of http requests to svn.wordpress.org. I think there is already an issue in the wpvulndb repo for fetching the current version and storing them

ethicalhack3r:
Good point, I think this makes more sense

ethicalhack3r:
Quick update: We have implemented current versions and release dates to plugins (and themes?) but not WP itself yet.

ethicalhack3r:
Just remembered that we already store the WP release date in WPVULNDB but don't export it to WPScan. Looking into this now. Current WP versions would be tricky.

ethicalhack3r:
We now show the WP release date

PR #929, opened 2016-04-20
WP Metadata Integration

ethicalhack3r:
https://github.com/wpscanteam/wpscan/issues/704


lib/common/models/wp_version/output.rb
```diff
@@ -4,8 +4,12 @@ class WpVersion < WpItem
   module Output
 
     def output(verbose = false)
+      metadata = self.metadata(self.number)
+
       puts
       puts info("WordPress version #{self.number} identified from #{self.found_from}")
+      puts " | Released: #{metadata[:release_date]}"
+      puts " | Changelog: #{metadata[:changelog_url]}"
```
FireFart: do we really need to show the changelog of the installed version? The interesting part would be the changelog of follow up versions, but we can't show this. So I would just put the release date in the Wordpress identified string.

ethicalhack3r: if verbose?

FireFart: jeah why not. But i would always print the release date since it's good metadata to get a feeling how old the installation is

lib/common/models/wp_version/output.rb
```diff
@@ -4,8 +4,12 @@ class WpVersion < WpItem
   module Output
 
     def output(verbose = false)
+      metadata = self.metadata(self.number)
+
       puts
       puts info("WordPress version #{self.number} identified from #{self.found_from}")
+      puts " | Released: #{metadata[:release_date]}"
+      puts " | Changelog: #{metadata[:changelog_url]}" if verbose
```
FireFart: I would do it this way to keep the output clear:

```
if verbose
    puts info("WordPress version #{self.number} identified from #{self.found_from}")
    puts " | Released: #{metadata[:release_date]}"
    puts " | Changelog: #{metadata[:changelog_url]}"
else
    puts info("WordPress version #{self.number} identified from #{self.found_from} #{" (Released on #{metadata[:release_date]})" if metadata[:release_date]}")
end
```

FireFart:
pinging @erwanlr to have a look at it (i think the parsing inside the output module is wrong) and to hopefully implement it in v3 too :)

PR #946, opened 2016-05-17
optimize png images losslessly using zopflipng

PeterDaveHello:
```
 sucuri_logo.png        | Bin 6037 -> 2617 bytes
 wpscan.png             | Bin 10671 -> 10093 bytes
 wpscan_logo_407x80.png | Bin 10394 -> 6904 bytes
```


FireFart:
@PeterDaveHello thx!

PR #949, opened 2016-05-31
more advanced version detection

FireFart:
Before: The first version hit returns the version
Now: All finders are executed and the version with the most matches is selected

### DIFF
--- a/lib/common/models/wp_item/findable.rb
+++ b/lib/common/models/wp_item/findable.rb
@@ -9,7 +9,7 @@ class WpItem
   #
   # @return [ void ]
   def found_from=(method)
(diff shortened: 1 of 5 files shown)


lib/common/models/wp_item/findable.rb
```diff
@@ -9,8 +9,7 @@ class WpItem
   #
   # @return [ void ]
   def found_from=(method)
-    found       = method[%r{find_from_(.*)}, 1]
-    @found_from = found.gsub('_', ' ') if found
+    @found_from = method.to_s.to_s.gsub(/find_from_/, '').gsub(/_/, ' ')
```
ethicalhack3r: ```.to_s``` twice, is this interntional?

FireFart: thanks fixed

Issue #841, opened 2015-07-02
error:  undefined method `cells' for #<Array:0x000000029cc2f8>


mytc:
When I try enumerate or brute force, I received this error 
```
undefined method `cells' for #<Array:0x000000029cc2f8>
```
Tried this on Majaro Arch and Centos 7.

whoknowwhat:
Linux 3.10.0-123.4.4.el7.x86_64 #1 SMP Fri Jul 25 05:07:12 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux
ruby 2.0.0p598 (2014-11-13) [x86_64-linux]
CentOS 7
WPScan Current version: 2.8
Last DB update: 2015-07-02

I got same message too.

ethicalhack3r:
Reproduced, problem seems to be terminal-table version 1.5.2. I will check to see if we can use an older version.

erwanlr:
use ```gem 'terminal-table', '~>1.4.5'``` in the Gemfile then ^^

ethicalhack3r:
Temporary fix is:

```
gem uninstall terminal-table
git pull
bundle install
```

FireFart:
We could also include our Gemfile.lock in the repo so the gem versions will always be the ones we are working with

mytc:
temporary fix by @ethicalhack3r  works.

Karanja-the-african-hippy:
Still getting this error "undefined method `cells' for #<Array:0x00000004a41c40>"
![screenshot](https://cloud.githubusercontent.com/assets/11226110/8868568/ab087b76-31e2-11e5-9625-513c7793c7cf.png)

ethicalhack3r:
What version of WPScan are you running? What version of the terminal-table gem?

Karanja-the-african-hippy:
WPScanis=2.8 terminal-table=1.5.2
![screenshot](https://cloud.githubusercontent.com/assets/11226110/8869578/0eca8acc-31ee-11e5-9136-f89165c1ca81.png)

ethicalhack3r:
Have you tried the temporary fix I mentioned above?

Karanja-the-african-hippy:
I don't know which repository to pull? in the git command

Karanja-the-african-hippy:
I changed my gem file but still shows same error ![screenshot](https://cloud.githubusercontent.com/assets/11226110/8899011/7aee02d8-3437-11e5-9608-1b29c0a2af88.png)

[...]

PR #950, opened 2016-06-06
bump terminal-table to 1.6.0 and drop workaround

anthraxx:
a new version of terminal-table has been released that makes the version pinning and workaround obsolete


ethicalhack3r:
Tested the PR locally, don't seem to get the error in https://github.com/wpscanteam/wpscan/issues/841 anymore. Thanks for the PR! :)

PR #973, opened 2016-08-05
Enable --cache-dir option in command line parameters, which solves write permission issues when wpscan is installed in system or root-owned directories

pierre-dargham:
## Overview :

This pull request just adds two lines to `wpscan_options.rb`, allowing users to pass a `--cache-dir` option from the command line. An example would be : `ruby wpscan.rb --url="http://example.com" --cache-dir="~/.wpscan/cache"`

## Further explanations :

Currently, wpscan writes the browser cache in its own directory `./cache/browser/`. This causes write permission errors when wpscan is installed in a system root directory such as `/usr/share/wpscan/` or `/usr/lib/wpscan/` and then executed by a user who doesn't have the permissions to write inside those directories.

Obviously, we don't want to grant write permission to non-root users in such directories, so we need to allow wpscan execution to write cache elsewhere.

Because the `browser.rb` Browser class takes its option array from the command line, we just need to allow a --cache-dir parameter to be passed when executing wpscan, which is what we added here.

### DIFF
--- a/lib/wpscan/wpscan_options.rb
+++ b/lib/wpscan/wpscan_options.rb
@@ -45,7 +45,8 @@ class WpscanOptions
     :no_banner,
     :throttle,
     :disable_accept_header,
-    :disable_referer
+    :disable_referer,
+    :cache_dir
   ]
 
   attr_accessor *ACCESSOR_OPTIONS
@@ -288,7 +289,8 @@ class WpscanOptions
       ['--no-banner', GetoptLong::NO_ARGUMENT],
       ['--throttle', GetoptLong::REQUIRED_ARGUMENT],
       ['--disable-accept-header', GetoptLong::NO_ARGUMENT],
-      ['--disable-referer', GetoptLong::NO_ARGUMENT]
+      ['--disable-referer', GetoptLong::NO_ARGUMENT],
+      ['--cache-dir', GetoptLong::REQUIRED_ARGUMENT]
     )
   end
 


ethicalhack3r:
Thanks for the pull request! Looking at your commit it only looks like the ```cache-dir``` option is added but nothing is done with?

pierre-dargham:
Thank you for reading this pull request,

I didn't have to change anything else to make it work, because the `Browser` class was already initialized this way :

 **`./lib/common/browser.rb line 35:36 :`**
 ```ruby
def initialize(options = {})
  @cache_dir   = options[:cache_dir] || CACHE_DIR + '/browser'
 ```

And the options array comes from command line options :

 **`./wpscan.rb line 15 :`**
 ```ruby
wpscan_options = WpscanOptions.load_from_arguments
```

 **`./wpscan.rb line 49:51 :`**
 ```ruby
Browser.instance(
  wpscan_options.to_h.merge(max_threads: wpscan_options.threads)
)
 ```

So the only issue was that `--cache-dir` was not allowed as a command line parameter. With this commit, the option seems to be properly used by the `Browser` class.

I made some basic testing where I successfully executed a system-installed wpscan as a non-root user. The tests passed without write permission issues, and the cache was effectively written in my `~/.wpscan/cache/` directory.

Some more advanced testing could be necessary, but I don't have much time to read the entire programm. I'm not a ruby developper, so I wanted to stay very simple whilst still making an improvement of a base use case by solving the write permission issue.

ethicalhack3r:
Thanks!

PR #976, opened 2016-08-09
Avoid false-positive detection of plugins and themes

l3msh0:
I found WPScan can misdetect too many non-existant plugins/themes under the following conditions.
- A target server returns `403` instead of `404` when requested plugins/themes URL doesn't exist
- The response bodies of such requests deffer from ordinary 404 response

This problem occurs when accesses to non-existant file under plugins/themes directories are explicitly prohibited by server settings.

On such situation, the server returns 403 response and its hash does not match ordinary 404 response, so `WpItem::Existable::exists_from_response?` can't tell whether a plugin exists or accesses are simply restricted. As a result, all plugins/themes are output as found.

### DIFF
--- a/lib/wpscan/wp_target.rb
+++ b/lib/wpscan/wp_target.rb
@@ -190,9 +190,9 @@ class WpTarget < WebSite
     unless @error_404_hash_set
       hash_set = Set.new([site_404_hash])
 
-      non_existant_page = Digest::MD5.hexdigest(rand(999_999_999).to_s) + '.html'
+      non_existent_page = Digest::MD5.hexdigest(rand(999_999_999).to_s) + '.html'
       [wp_content_dir, wp_plugins_dir, wp_themes_dir].each do |dir|
-        hash_set << WebSite.page_hash(@uri.merge("#{dir}/#{non_existant_page}"))
+        hash_set << WebSite.page_hash(@uri.merge("#{dir}/#{non_existent_page}"))
       end
       @error_404_hash_set = hash_set.freeze
     end
--- a/lib/wpscan/wp_target.rb
+++ b/lib/wpscan/wp_target.rb
@@ -1,5 +1,6 @@
 # encoding: UTF-8
 
+require 'set'
 require 'web_site'
 require 'wp_target/wp_readme'
 require 'wp_target/wp_registrable'
@@ -176,5 +177,25 @@ class WpTarget < WebSite
 
   def include_directory_listing_enabled?
     directory_listing_enabled?(includes_dir_url)
+  end
+
+  # Return the set of 404 error page hash for wp-content directories.
+  #
+  # This set is useful to check if a plugin or a theme is installed
+  # because 404 error pages for non-existent items can be different from
+  # target's default because of some plugin(i.e. IP Geo Block).
+  #
+  # @return [ Set<String> ]
+  def error_404_hash_set
+    unless @error_404_hash_set
+      hash_set = Set.new([site_404_hash])
+
+      non_existant_page = Digest::MD5.hexdigest(rand(999_999_999).to_s) + '.html'
+      [wp_content_dir, wp_plugins_dir, wp_themes_dir].each do |dir|
+        hash_set << WebSite.page_hash(@uri.merge("#{dir}/#{non_existant_page}"))
+      end
+      @error_404_hash_set = hash_set.freeze
+    end
+    @error_404_hash_set
   end
 end
--- a/lib/wpscan/wp_target/wp_custom_directories.rb
+++ b/lib/wpscan/wp_target/wp_custom_directories.rb
@@ -26,7 +26,7 @@ class WpTarget < WebSite
 
       if WpTarget.valid_response_codes.include?(response.code)
         hash = WebSite.page_hash(response)
-        return true if hash != error_404_hash and hash != homepage_hash
+        return true if hash != site_404_hash and hash != homepage_hash
       end
 
       false
@@ -43,6 +43,14 @@ class WpTarget < WebSite
     # @return [ Boolean ]
     def wp_plugins_dir_exists?
       Browser.get(@uri.merge(wp_plugins_dir).to_s).code != 404
+    end
+
+    # @return [ String ] The wp-themes directory
+    def wp_themes_dir
+      unless @wp_themes_dir
+        @wp_themes_dir = "#{wp_content_dir}/themes"
+      end
+      @wp_themes_dir
     end
 
   end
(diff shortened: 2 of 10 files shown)


l3msh0:
This problem occurs when target site uses [IP Geo Block](https://wordpress.org/plugins/ip-geo-block/) plugin. It returns 403 response against accesses to plugins/themes directory.

ethicalhack3r:
Look OK to you @FireFart @erwanlr? Looks good to me but not tested it in the real world.

erwanlr:
Can't the --exclude-content-based option be used in this case ? It was
designed for it

On 12 Aug 2016 11:15, "Ryan Dewhurst" <notifications@github.com> wrote:

Look OK to you @FireFart <https://github.com/FireFart> @erwanlr
<https://github.com/erwanlr>? Looks good to me but not tested it in the
real world.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/wpscanteam/wpscan/pull/976#issuecomment-239411250>, or mute
the thread

FireFart:
@l3msh0 @ethicalhack3r @erwanlr If a plugin 404 gives hash `x` and target 404 a `y`, this would not fire if the plugin dir for some reason returns hash `y` (the hash from the target). I think it would be better to create a hash list of known locations like uploads, themes, plugins, ... store them in a list, and check if the list contains the 404_hash.

l3msh0:
@erwanlr As you pointed out, '--exluce-content-based' option is usable for this case so I don't think this PR have to be merged.
Though I think it is not so easy for users to understand the cause of false-positive detection and retry with the properly option.

l3msh0:
@FireFart Making a hash list sounds better. I don't have confidence but should I implement it?

FireFart:
@l3msh0 sure try it :)

l3msh0:
OK. I'll try it 👌 
It may take a little time.

l3msh0:
Hi. It's been a long time.

I rewrote the code according to [@FireFart 's suggestion](https://github.com/wpscanteam/wpscan/pull/976#issuecomment-239533816).

`--exclude-content-based` option can be used in this case but I think it would be better that this PR was merged, because it takes too much time if all plugins and themes are misdetected. It makes WPScan look hung-up.

lib/wpscan/wp_target.rb
```diff
@@ -177,4 +178,24 @@ def upload_directory_listing_enabled?
...
+  # because 404 error pages for non-existent items can be different from
+  # target's default because of some plugin(i.e. IP Geo Block).
+  #
+  # @return [ Set<String> ]
+  def error_404_hash_set
+    unless @error_404_hash_set
+      hash_set = Set.new([site_404_hash])
+
+      non_existant_page = Digest::MD5.hexdigest(rand(999_999_999).to_s) + '.html'
+      [wp_content_dir, wp_plugins_dir, wp_themes_dir].each do |dir|
```
FireFart: All these directories contain blank index.php files for example. A lot of sites render a blank page when calling these directories so this would produce a lot of false positives. So we need some more checks in this place.

Also you are copying the `site_404_hash` method from above. Please use a single method for this so we do not duplicate code.

lib/wpscan/wp_target/wp_custom_directories.rb
```diff
@@ -26,7 +26,7 @@ def default_wp_content_dir_exists?
 
       if WpTarget.valid_response_codes.include?(response.code)
         hash = WebSite.page_hash(response)
-        return true if hash != error_404_hash and hash != homepage_hash
+        return true if hash != site_404_hash and hash != homepage_hash
```
FireFart: why don't you check against the hash set here?

lib/wpscan/wp_target/wp_must_use_plugins.rb
```diff
@@ -10,7 +10,7 @@ def has_must_use_plugins?
 
       if response && [200, 401, 403].include?(response.code)
         hash = WebSite.page_hash(response)
-        return true if hash != error_404_hash && hash != homepage_hash
+        return true if hash != site_404_hash && hash != homepage_hash
```
FireFart: why don't you check against the hash set here?

lib/wpscan/wp_target.rb
l3msh0: Sorry, I don't understand why you think blank index page brings false-positive.

Suppose a target returns following responses against each URL,

```
- #{random-md5-digest}.html                       -> 404                 hash X
- /wp-content/plugins/                            -> 200                 hash Y
- /wp-content/plugins/#{random-md5-digest}.html   -> 200 (but not found) hash Z
```

response hashes for other non-existent URLs under the plugin directory are expected to be Z (or maybe X) and this PR intends to treat them as "not found".

Whether directory index is blank page or not (= hash Y's value) has nothing to do with item detection, does it?

l3msh0: > Also you are copying the site_404_hash method from above. Please use a single method for this so we do not duplicate code.

`wp_conent_dir` method needs site's 404 page hash via `default_wp_content_dir_exists?` method, but `error_404_hash_set` cannot be used because it calls `wp_content_dir` and raises stackoverflow.
Therefore, `site_404_hash` method must be separated.

lib/wpscan/wp_target/wp_custom_directories.rb
l3msh0: As I mentioned above, calling `error_404_hash_set` method here leads to stackoverflow.

lib/wpscan/wp_target/wp_must_use_plugins.rb
l3msh0: OK. I've changed it to use the existent method.

lib/wpscan/wp_target.rb
FireFart: The problem is we check for wp-themes, wp-plugins and so on several times and I think if we qualify this hast as a not found hash, this will break some stuff :/

FireFart: Also if one of the directories has directory listing enabled, this would be treated like a 404.

l3msh0: @FireFart  I think the idea that referring the common hash set for any type of detection brings complexity and risk of breaking other parts.

It seems to be better that plugin's 404 hash would be used only for plugin detection (my first approach).
The same thing can be said about themes.

I've tried it on [another branch](https://github.com/wpscanteam/wpscan/compare/master...l3msh0:avoid-false-positive-detection-for-each-item-type). Please check it out.

>  If a plugin 404 gives hash x and target 404 a y, this would not fire if the plugin dir for some reason returns hash y (the hash from the target).

This problem remains but I don't think it is so important compared to the cost of introducing the hash set.

l3msh0: @FireFart ping

ethicalhack3r:
No longer applicable with new codebase

Issue #1018, opened 2016-10-24
Soft 404's are detected as plugin/theme installed


uglyrobot:
Scanning my domain returns every single vulnerable plugin/theme in the db for my site. For example https://andycrone.com/wp-content/plugins/redirection/

Shouldn't a 404 detect as not installed?

FireFart:
I think this is caused by cloudflare. We need to look into this, maybe we can detect smth like "SmartErrors powered by CloudFlare"

uglyrobot:
I've run into this with other sites not running Cloudflare too though, I think it just depends on the rewrite config if it gives a soft 404.

FireFart:
@uglyrobot could you please test with this PR?
https://github.com/wpscanteam/wpscan/pull/1019

(diff file: https://github.com/wpscanteam/wpscan/pull/1019.diff)

This patch removes script tags before calculating hashes like the error hash

ethicalhack3r:
did the patch work for you @uglyrobot ?

ethicalhack3r:
Closing due to lack of response

PR #1019, opened 2016-10-25
remove scripts before calculating hashes

FireFart:
This needs some testing, but may improve detection rates

See #1018 

CC @ethicalhack3r @erwanlr


ethicalhack3r:
Looks OK to me!

PR #1025, opened 2016-11-22
Alpine based smaller docker image

dvestal:
I am not sure whether it is desired, but if so, here is a docker build based on Alpine rather than the ruby:2.3-slim image. It finishes at 92.35MB instead of 573.7MB.

The biggest question that I have with this is why the wpscan user/group were created in the original image. I couldn't find anything that doesn't worked because of running things as root instead, but I can add that back in if necessary.

screen shot 2016-11-22 at 8 18 16 am

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,23 +1,17 @@
-FROM ruby:2.3-slim
+FROM alpine:3.4
 MAINTAINER WPScan Team <team@wpscan.org>
 
-RUN DEBIAN_FRONTEND=noninteractive && \
-  rm -rf /var/lib/apt/lists/* && \
-  apt-get update && \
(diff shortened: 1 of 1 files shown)


FireFart:
@dvestal thanks I will have a look later. We run as non root user because there is no need to run as root. It's a bad behaviour all docker images run as root by default .......

FireFart:
@dvestal We currently have a hook on the ruby base image on docker hub so the image is rebuild every time the base image changes (eg a new version of ruby is released). We loose this ability with the use of the alpine image because ruby is installed via apk. This would cause an overhead on our side because we have to manually rebuild the image each time instead of being automatically updated.
I will have a look into using the official ruby alpine image instead and see if this can reduce the image size.

But thanks for your PR!

FireFart:
@dvestal I updated the docker image to alpine:
https://github.com/wpscanteam/wpscan/commit/67ba526b5b2debbcab1988305307a1fbebb7b618
Size got down from 551MB to 168MB! Thx for the hint!

PR #1045, opened 2017-01-11
Add gcc to Debian prerequisites

thijskh:
This is needed to install some gems and mirrors the fact that gcc is included in the command lines Fedora and Ubuntu (there contained in `build-essential`).


FireFart:
@thijskh thanks!

Issue #854, opened 2015-08-17
Finding version # from changelogs


brandonsturgeon:
This is _sort of_ a 2-part issue. 1 being that we should have aliases for changelog filenames, and the other being that if we can't find a version number from a readme.txt, we could look in the changelog.

Looks like scans are only looking for `changelog.txt` but some plugins (eg: gravity forms) have a `change_log.txt` instead.

Here's where it's checking, it looks like. 
https://github.com/wpscanteam/wpscan/blob/master/lib/common/models/wp_item/infos.rb#L39

We can do something similar to how we handled the `readme.txt` situation slightly above 
https://github.com/wpscanteam/wpscan/blob/master/lib/common/models/wp_item/infos.rb#L14

Here's what a scan looks like while on master

```
_______________________________________________________________
        __          _______   _____
        \ \        / /  __ \ / ____|
         \ \  /\  / /| |__) | (___   ___  __ _ _ __
          \ \/  \/ / |  ___/ \___ \ / __|/ _` | '_ \
           \  /\  /  | |     ____) | (__| (_| | | | |
            \/  \/   |_|    |_____/ \___|\__,_|_| |_|

        WordPress Security Scanner by the WPScan Team
                       Version 2.8
          Sponsored by Sucuri - https://sucuri.net
   @_WPScan_, @ethicalhack3r, @erwan_lr, pvdl, @_FireFart_
_______________________________________________________________

[+] URL: http://www.SITE.com/
[+] Started: Mon Aug 17 11:31:54 2015

[+] Interesting header: LINK: <http://www.SITE.com/>; rel=shortlink
[+] Interesting header: SERVER: nginx/1.8.0
[+] Interesting header: X-POWERED-BY: PHP/5.3.3
[+] XML-RPC Interface available under: http://www.SITE.com/xmlrpc.php

[+] WordPress version 4.2.4 identified from meta generator

[+] WordPress theme in use: twentyfifteen - v1.1

[+] Name: twentyfifteen - v1.1
 |  Location: http://www.SITE.com/wp-content/themes/twentyfifteen/
 |  Readme: http://www.SITE.com/wp-content/themes/twentyfifteen/readme.txt
 |  Style URL: http://www.SITE.com/wp-content/themes/twentyfifteen/style.css
 |  Theme Name: Twenty Fifteen
 |  Theme URI: https://wordpress.org/themes/twentyfifteen/
[...]

PR #1071, opened 2017-03-16
Improve plugin version detection

errm:
Some plugins we are using don't have a readme, but the version can be
detected from a changelog.

### DIFF
--- a/lib/common/models/wp_item/infos.rb
+++ b/lib/common/models/wp_item/infos.rb
@@ -14,11 +14,7 @@ class WpItem
     def readme_url
       # See https://github.com/wpscanteam/wpscan/pull/737#issuecomment-66375445
       # for any question about the order
-      %w{readme.txt README.txt Readme.txt ReadMe.txt README.TXT readme.TXT}.each do |readme|
-        url = @uri.merge(readme).to_s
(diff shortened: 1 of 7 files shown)


ethicalhack3r:
Thanks for the pull request!

Do you know how popular the additional changelog files you've added are in real life? ```changes.txt changelog.md```

You could find this out by downloading all Plugins from the SVN repository (```http://plugins.svn.wordpress.org/```) and doing a search for the changelog files you've added. (this will take a while)

My concern is increasing the number of HTTP requests unnecessarily, if the benefit of the additional checks is low. However, if the code stops once it has a successful version match (doesnt check for the other changelog files), then I think this would lower the number of requests overall.

@erwanlr @FireFart what do you think?

Another issue is that this version of WPScan is on its way out and will be replaced by the new branch in the future (https://github.com/wpscanteam/wpscan-v3) so not sure if we should add this to WPScan v3 instead, if we decide it is beneficial.

erwanlr:
As mentioned here: https://github.com/wpscanteam/wpscan/issues/854#issuecomment-132345774

changes.txt is definitely something I never saw been used. changelog.md is barely used from the research I did a year ago, could re-run the script to see

PR #1075, opened 2017-03-25
Preciser reporting in bruteforcing password with bad response

qutorial:
When bruteforcing for multiple logins and passwords the bad response code reported
might indicate a match! But the reporting for it is not clear enough.
For example "Unkown response for admin" might mean a user name admin and some password
or a password 'admin' for some user.

This commit makes in unambiguous reporting a bad response, and naming which login and
which password caused it.


erwanlr:
Thank you!

PR #1080, opened 2017-04-07
allows wordlist to take command

jamesalbert:
This is my first time writing anything ruby, so please let me know if this is not the ruby way of doing things and tell me to go to hell. I figured the best use case for this would be providing a crunch command like so:

`wpscan.rb --url <target> --username <user> --wordlist "crunch 10 13 -f charset.lst hex-lower"`

edit: I initially tried to add the feature so that you could do `crunch 10 13 ... | wpscan.rb ...`, but it was only reading the first line. Maybe there's a ruby way of working around that, and if so let me know, but I just implemented the feature this way instead.

### DIFF
--- a/lib/common/models/wp_user/brute_forcable.rb
+++ b/lib/common/models/wp_user/brute_forcable.rb
@@ -28,9 +28,19 @@ class WpUser < WpItem
       queue_count  = 0
       found        = false
 
-      create_progress_bar(count_file_lines(wordlist)+1, options)
+      if File.file?(wordlist)
+        wordlist = File.readlines(wordlist)
+      else
+        command = wordlist
+        wordlist = []
+        IO.popen(command).each do |line|
+          wordlist << line.chomp
+        end
+      end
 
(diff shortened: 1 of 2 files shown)


ethicalhack3r:
Thanks for the PR. Unfortunatly I don't think this is something we would want in the main repository. It opens up the risk of command execution, where the benefit is not huge (could run crunch separately I think?).

FireFart:
@ethicalhack3r @jamesalbert I think the correct way to implement this would be to do smth like this:
```
cat wordlist | wpscan -u .... --username --wordlist -
```
We don't want wpscan to execute commands on the users computer

FireFart:
@ethicalhack3r to be clear: this is currently not implemented. So someone needs to implement `-` as stdin.

@erwanlr is this possible in v3?

jamesalbert:
wow, can't believe I submitted this pr. Didn't really think that one through... Implemented `--wordlist -` to read stdin as suggested in #1081

PR #1081, opened 2017-04-10
--wordlist - reads stdin

jamesalbert:


lib/common/models/wp_user/brute_forcable.rb
```diff
@@ -28,9 +28,15 @@ def brute_force(wordlist, options = {}, redirect_url = nil)
       queue_count  = 0
       found        = false
 
-      create_progress_bar(count_file_lines(wordlist)+1, options)
+      if wordlist == '-'
+        wordlist = $stdin.readlines
```
FireFart: this would read the wordlist completely into memory (which can be bad in case of a big wordlist)

I think we need to pass this info to `brute_force` which should then loop over stdin line by line (or the wordlist)

jamesalbert: updated it to read line by line. would something like that do the job?

FireFart:
@ethicalhack3r would you mind giving this one a quick test? Looks good so far

ethicalhack3r:
It doesn't show the total number of requests, progress precentage or the ETA in the brute force progress bar.

Command:
```
cat ~/Tools/wordlists/500-worst-passwords.txt | ./wpscan.rb -u www.example.com --wordlist -
```

Here is the output:
```
 Brute Forcing 'admin' Time: 00:00:13 <--=---=---> (226 / ) 0.0%  ETA: ??:??:??
```

jamesalbert:
Is there a way to tell how many lines you're getting from stdin without reading it all into memory?

FireFart:
@jamesalbert nope because you are consuming stdin line per line to be efficient. Is there a way we can only display guesses so far and hide the percentage and ETA?

FireFart:
@jamesalbert https://github.com/jfelchner/ruby-progressbar/wiki/Unknown-Progress

ethicalhack3r:
Here is how we count lines currently - https://github.com/wpscanteam/wpscan/blob/master/lib/common/common_helper.rb#L234

FireFart:
@ethicalhack3r but this does not make sense when dealing with stdin :)

jamesalbert:
let me know if there's any thing else that doesn't seem legit.

jamesalbert:
that should do it

jamesalbert:
the deed is done 👍

lib/common/models/wp_user/brute_forcable.rb
```diff
@@ -52,16 +52,19 @@ def brute_force(wordlist, options = {}, redirect_url = nil)
         request = login_request(password, redirect_url)
 
         request.on_complete do |response|
-          progress_bar.progress += 1 if options[:show_progression] && !found
+          if options[:show_progression] && !found
+            progress_bar.progress += 1
+            if progress_bar.progress.fdiv(progress_bar.total) >= 0.8
+              progress_bar.total *= 2
+            end
```
ethicalhack3r: This could be simplified with:

```progress_bar.total *= 2 if progress_bar.progress.fdiv(progress_bar.total) >= 0.8```

jamesalbert: true, true. but is breaking the 80 character limit worth it??

ethicalhack3r: Ah! You're right, nah not worth it.

ethicalhack3r:
Looks good to me, thanks! :) Can you do one last change please? Just mention the new "-" option in the README.md file.

Something like:
```--wordlist | -w <wordlist>          Supply a wordlist for the password brute forcer. If the "-" option is supplied, the wordlist is expected via STDIN.```

jamesalbert:
No worries, this is a cool project and I'm happy to contribute. Sounds good man, I would have done it sooner but it was 4am where I am and I knocked out.

and hey, let's just keep #1080 between us. Nobody has to know...

ethicalhack3r:
Thanks for your work on this! :)

FireFart:
@jamesalbert awesome thanks for your work on this!

PR #1097, opened 2017-05-26
silly logging hack

skout23:
ugly hack to allow passing a file for logging, instead of having a default file of log.txt in the scans running directory.


skout23:
sorry meant to make this PR on my local fork only.

PR #1109, opened 2017-07-07
Document Homebrew package in README

zmwangx:
[Homebrew](https://brew.sh/) is arguably the most popular package manager for macOS.


FireFart:
thx!

PR #1164, opened 2018-01-31
Don't use colour codes when stdout is redirected

rbsec:
When output is redirected (either to a file or piped to another program), the colour codes are normally a pain and have to be filtered out (as per the instructions on the wiki).

We can easily detect if stdout is a tty, and just suppress the colour codes if it's not.

### DIFF
--- a/lib/common/common_helper.rb
+++ b/lib/common/common_helper.rb
@@ -103,7 +103,7 @@ end
 
 # Define colors
 def colorize(text, color_code)
-  if $COLORSWITCH
+  if $COLORSWITCH or not $stdout.isatty
     "#{text}"
   else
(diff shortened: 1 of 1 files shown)


erwanlr:
Isn't the ```--no-color``` suitable for what you are after ?

rbsec:
Erwanlr,

Yes, `--no-color` provides this functionality, but whenever I'm redirecting the output to somewhere I always want that option (and sometimes forget to do it, which is a pain), so it seemed more sensible to have that as a default.

Feel free to close this if you'd prefer to keep it as it.

~rbsec

erwanlr:
I will close it then.

PR #1179, opened 2018-04-23
Added 'Search-Replace-DB' from Interconnect IT to the list of paths to l…

roberts91:
Hello!

This PR adds an extra line in to list of paths to look for exposed config files in.
The line is not a config file itself, but is is the path to a folder called "Search-Replace-DB".

[Search-Replace-DB](https://github.com/interconnectit/Search-Replace-DB) is a popular migration script for WordPress sites. The issue with this is that people tend to keep this script on the server, not cleaning properly after a migration. The even bigger issue is that this script reads the DB credentials from the wp-config.php-file if it is found either in the parent or the second parent folder above the Search-Replace-DB-script.

I think it would be useful for WPScan to look for this. Do you guys agree?


ethicalhack3r:
No longer valid with new codebase

PR #1182, opened 2018-05-11
Multiple Features

g0tmi1k:
- Add off-line database update support
- Add a few more checks/fail safes
- Add `/humans.txt` check
- Add `/.well-known/security.txt` check
- Switch user folder location to `~/.wpscan` (useful for users who are not root and use a package)
- Grammar fixes
- Output is easier to understand
- More verbose messages

### DIFF
--- a/lib/common/common_helper.rb
+++ b/lib/common/common_helper.rb
@@ -1,34 +1,35 @@
 # encoding: UTF-8
 
-LIB_DIR              = File.expand_path(File.join(__dir__, '..'))
-ROOT_DIR             = File.expand_path(File.join(LIB_DIR, '..')) # expand_path is used to get "wpscan/" instead of "wpscan/lib/../"
-DATA_DIR             = File.join(ROOT_DIR, 'data')
-CONF_DIR             = File.join(ROOT_DIR, 'conf')
-CACHE_DIR            = File.join(ROOT_DIR, 'cache')
-WPSCAN_LIB_DIR       = File.join(LIB_DIR, 'wpscan')
-UPDATER_LIB_DIR      = File.join(LIB_DIR, 'updater')
-COMMON_LIB_DIR       = File.join(LIB_DIR, 'common')
-MODELS_LIB_DIR       = File.join(COMMON_LIB_DIR, 'models')
-COLLECTIONS_LIB_DIR  = File.join(COMMON_LIB_DIR, 'collections')
+LIB_DIR              = File.expand_path(File.join(__dir__, '..')) # wpscan/lib/
+ROOT_DIR             = File.expand_path(File.join(LIB_DIR, '..')) # wpscan/ - expand_path is used to get "wpscan/" instead of "wpscan/lib/../"
+USER_DIR             = File.expand_path(Dir.home) # ~/
+CACHE_DIR            = File.join(USER_DIR, '.wpscan/cache') # ~/.wpscan/cache/
+DATA_DIR             = File.join(USER_DIR, '.wpscan/data') # ~/.wpscan/data/
+CONF_DIR             = File.join(ROOT_DIR, 'conf') # wpscan/conf/
+COMMON_LIB_DIR       = File.join(LIB_DIR, 'common') # wpscan/lib/common/
+UPDATER_LIB_DIR      = File.join(LIB_DIR, 'updater') # wpscan/lib/updater/ -  Not used ATM
+WPSCAN_LIB_DIR       = File.join(LIB_DIR, 'wpscan') # wpscan/lib/wpscan/
+COLLECTIONS_LIB_DIR  = File.join(COMMON_LIB_DIR, 'collections') # wpscan/lib/common/collections/ -  Not used ATM
+MODELS_LIB_DIR       = File.join(COMMON_LIB_DIR, 'models') # wpscan/lib/common/models/
 
-DEFAULT_LOG_FILE     = File.join(ROOT_DIR, 'log.txt')
+DEFAULT_LOG_FILE     = File.join(USER_DIR, '.wpscan/log.txt')  # ~/.wpscan/log.txt
 DATA_FILE            = File.join(ROOT_DIR, 'data.zip') # wpscan/data.zip
 
 # Plugins directories
-COMMON_PLUGINS_DIR   = File.join(COMMON_LIB_DIR, 'plugins')
-WPSCAN_PLUGINS_DIR   = File.join(WPSCAN_LIB_DIR, 'plugins') # Not used ATM
+COMMON_PLUGINS_DIR   = File.join(COMMON_LIB_DIR, 'plugins') # wpscan/lib/common/plugins/ -  Not used ATM
+WPSCAN_PLUGINS_DIR   = File.join(WPSCAN_LIB_DIR, 'plugins') # wpscan/lib/common/plugins/ -  Not used ATM
 
-# Data files
-WORDPRESSES_FILE  = File.join(DATA_DIR, 'wordpresses.json')
-PLUGINS_FILE      = File.join(DATA_DIR, 'plugins.json')
-THEMES_FILE       = File.join(DATA_DIR, 'themes.json')
-TIMTHUMBS_FILE    = File.join(DATA_DIR, 'timthumbs.txt')
-WP_VERSIONS_FILE  = File.join(DATA_DIR, 'wp_versions.xml')
-LOCAL_FILES_FILE  = File.join(DATA_DIR, 'local_vulnerable_files.xml')
-WP_VERSIONS_XSD   = File.join(DATA_DIR, 'wp_versions.xsd')
-LOCAL_FILES_XSD   = File.join(DATA_DIR, 'local_vulnerable_files.xsd')
-USER_AGENTS_FILE  = File.join(DATA_DIR, 'user-agents.txt')
-LAST_UPDATE_FILE  = File.join(DATA_DIR, '.last_update')
+# Data files (data.zip)
+LAST_UPDATE_FILE    = File.join(DATA_DIR, '.last_update') # ~/.wpscan/data/.last_update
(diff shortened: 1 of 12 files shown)


lib/common/common_helper.rb
```diff
@@ -1,25 +1,29 @@
...
+# Location directories
 LIB_DIR              = File.expand_path(File.join(__dir__, '..')) # wpscan/lib/
 ROOT_DIR             = File.expand_path(File.join(LIB_DIR, '..')) # wpscan/ - expand_path is used to get "wpscan/" instead of "wpscan/lib/../"
 USER_DIR             = File.expand_path(Dir.home) # ~/
+
+# Core WPScan directories
 CACHE_DIR            = File.join(USER_DIR, '.wpscan/cache') # ~/.wpscan/cache/
 DATA_DIR             = File.join(USER_DIR, '.wpscan/data') # ~/.wpscan/data/
-CONF_DIR             = File.join(ROOT_DIR, 'conf') # wpscan/conf/
+CONF_DIR             = File.join(ROOT_DIR, '.wpscan/conf') # ~/.wpscan/conf/
```
ethicalhack3r: I think the path is missing the ```~```? As it currently is, wouldn't this create a hidden directory within the wpscan directory, rather than the user's home folder?

ethicalhack3r:
Hey @g0tmi1k! Thanks for the PR, lots of good cleanup and new stuff here!

Just a couple of points from me, maybe @erwanlr and @FireFart have some feedback also.

For me, ```Add /humans.txt check```and ```Add /.well-known/security.txt check``` are not WordPress specific enough. We have to be careful we don't become another nikto, which already does an awesome job in this regard. I would not implement these checks.

For the offline database extraction, I think it would only be needed if the user has no Internet access, and the database files shipped with wpscan are not yet unzipped? Are there any other scenarios that I may not be thinking of?

g0tmi1k:
Thanks @ethicalhack3r !

Been meaning todo some of this for a while...

I would like you to reconsider `humans.txt`/`security.txt` checks.
Personally, I would like to see wpscan a "One Stop Shop" regarding (enumeration of) WordPress installs. I don't want to be running another tool to get some other (basic/simple) information.
In regards to nikto, its doing above and beyond this. As a result, as nikto currently stands, the output is un-needed, clutted and often litted with false positive (as a result, I don't often use this in my work flow). But I see your point, there is no need to add other tools functionally into wpscan (such as sslscan - which I run before wpscan). Who knows. Maybe one day there can be some integrations into other tools.
As you are already checking for other things in wpscan (such as headers and `robots.txt` which do not link to wordpress's core), adding a few more standard checks files would be a benefit.
Could @FireFart / @erwanlr weigh-in on the matter too?

I agree on the offline support check.
Will update shortly.

g0tmi1k:
The 8x issues with codeclimate.com, are they required for the PR to be accepted @ethicalhack3r ?

FireFart:
wow great PR! :thumbsup:
I get both points of not reinventing the wheel but it's hard to say where to stop (for example should we also try all possible HTTP verbs?)
I would leave the new functions as they do not add much overhead in traffic during scan and add a benefit for the user.
I will have a look at the other changes the next days. Maybe the docker file also needs some update because of the changed location of the data folder.

PS: I think the most useful feature for v2 would be to skip the passive detection part and go straight to bruteforcing or whatever, but that's for a seperate PR I think :)

FireFart:
oh and travis is failing because the gemfile.lock does not match. Can you please run a bundle update and commit the changes to the lockfile?

g0tmi1k:
Thanks @FireFart 
I've just checked docker this morning, and I didn't run into any issues using it.

And I agree, skipping the "intro" checks would be a lovely idea. Is that something for WPscan v2 or v3 - _I've just seen (or re-remembered) there is WPScan v3 repo ( which is beta/WIP)_.

Ahhhh! Thanks for the nudge regarding `gemfile.lock`. Should be sorted now =).

lib/common/common_helper.rb
g0tmi1k: I agree!

lib/common/common_helper.rb
```diff
@@ -78,13 +80,39 @@ def add_trailing_slash(url)
...
+def missing_db_files?
   DbUpdater::FILES.each do |db_file|
     return true unless File.exist?(File.join(DATA_DIR, db_file))
   end
   false
 end
 
+# Find data.zip?
+def has_db_zip?
+  return File.exist?(DATA_FILE)? true : false
```
FireFart: can just do `return File.exist?(DATA_FILE)` :)

lib/common/common_helper.rb
```diff
@@ -78,13 +80,39 @@ def add_trailing_slash(url)
...
+# Extract data.zip
+def extract_db_zip
+  Zip::File.open(DATA_FILE) do |zip_file|
+    zip_file.each do |f|
+      # Feedback to the user
+      #puts "[+] Extracting: #{File.basename(f.name)}"
+      f_path = File.join(DATA_DIR, File.basename(f.name))
+
+      # Create folder
+      FileUtils.mkdir_p(File.dirname(f_path))
```
FireFart: can we move this up to the top and just create the `DATA_DIR`? There are currently no plans in having subfolders in the data directory

lib/wpscan/web_site/sitemap.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      # Is there a sitemap value?
+      result = response.body.scan(/^sitemap\s*:\s*(.*)$/i)
+      return true if result[0]
+      return false
+    end
+
+    # Gets a robots.txt URL
+    # @return [ String ]
+    def sitemap_url
```
FireFart: I think this should be renamed or return the parsed sitemap_url (logic from above)

.dockerignore
```diff
@@ -1,21 +1,6 @@
...
 .*
-**/*.md
-*.md
-Dockerfile
-**/*.orig
-*.orig
-CREDITS
-data.zip
-DISCLAIMER.txt
-example.conf.json
```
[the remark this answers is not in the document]
g0tmi1k: Fair enough. Will add back in.

lib/common/common_helper.rb
```diff
@@ -1,32 +1,33 @@
...
+WPSCAN_LIB_DIR       = File.join(LIB_DIR, 'wpscan') # wpscan/lib/wpscan/
+MODELS_LIB_DIR       = File.join(COMMON_LIB_DIR, 'models') # wpscan/lib/common/models/
+
+# Core WPScan files
+DEFAULT_LOG_FILE     = File.join(USER_DIR, '.wpscan/log.txt')  # ~/.wpscan/log.txt
+DATA_FILE            = File.join(ROOT_DIR, 'data.zip') # wpscan/data.zip
+
+# WPScan Data files (data.zip)
+LAST_UPDATE_FILE    = File.join(DATA_DIR, '.last_update') # ~/.wpscan/data/.last_update
+LOCAL_FILES_FILE    = File.join(DATA_DIR, 'local_vulnerable_files.xml') # ~/.wpscan/data/local_vulnerable_files.xml -  Not ref ATM
```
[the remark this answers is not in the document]
g0tmi1k: `LOCAL_FILES_FILE`, `LOCAL_FILES_XSD` and `WP_VERSIONS_XSD` are defined, but they are not called anywhere.
However, they are still in `data.zip` and used when updating (the filenames, not variables).

Will remove

lib/common/common_helper.rb
g0tmi1k: So you can! Thanks!

lib/common/common_helper.rb
g0tmi1k: Fair enough.
Was trying to future proof the code, but guess it makes it more complex.

Altered

lib/wpscan/web_site/sitemap.rb
g0tmi1k: Have you got a suggestion for the function name then?
I was trying to go on other names/standards in the code.

g0tmi1k:
Thanks @FireFart ! 
Commit applied based on your suggestions <3

lib/wpscan/web_site/sitemap.rb
FireFart: ok then let's just leave it. Thanks!

g0tmi1k:
🎉 
Thanks for the help @FireFart and @ethicalhack3r  <3

Keep up the good work @wpscanteam

g0tmi1k:
@FireFart Is it worth me making any more PRs to this repo? Or start with WPScan v3?

FireFart:
@g0tmi1k I think Wpscan v2 will be still around some time but v3 is the future. Alltough v3 still needs some more testing before we can release it. But Wpscan v3 should have better detection rates as it uses an advanced database :)

g0tmi1k:
If you had to take a guess, could you see v3 being released by the end of the year?

The last release of v2 was nearly a year ago (July 2017). Could there be an update for that soon?

FireFart:
We can release an update to v2 anytime. There was no update yet because there were no big changes to the codebase but now we should consider it (cc @ethicalhack3r)

On v3:
I think there were only a few testers so far of version 3 so we would need to get some more (mostly comparison of detection results). Also there may be some more potential in the usability because v3 has a lot of options that can be confusing to new users so we would love to have a feedback on that as well. Oh and last but not least the kali package will also need an overhaul because v3 is now a single gem. So it would be possible to release it soon but there is still a lot work to do :)

kimocoder:
Awesome work you did here @g0tmi1k will be happy to check it out

ethicalhack3r:
@FireFart let's release a new minor version of v2 in the next few weeks. Let's say the 8th June?

FireFart:
sounds good!

g0tmi1k:
Related: #1187

PR #1211, opened 2018-09-16
Fixing error on missformated rss

sudoaza:
When there's a link saying a file is a rss but its actually not nokogiri gives an exception. Small patch to catch it an return false. No regression spec included :(

### DIFF
--- a/lib/wpscan/wp_target/wp_rss.rb
+++ b/lib/wpscan/wp_target/wp_rss.rb
@@ -36,8 +36,13 @@ class WpTarget < WebSite
       # If there is nothing there, return false
       return false if data.empty?
 
-      # Read in RSS/XML
-      xml = Nokogiri::XML(data)
+      begin
+        # Read in RSS/XML
+        xml = Nokogiri::XML(data)
+      rescue
+        puts critical("Missformed XML")
+        return false
+      end
 
       begin
         # Look for <dc:creator> item


ethicalhack3r:
Thank you for the PR! Although, this branch of WPScan will be deprecated on September 26th, and replaced with WPScan v3.x (https://github.com/wpscanteam/wpscan-v3)

Do you also get the exception on WPScan v3.x?

FireFart:
@ethicalhack3r but we should merge it as v2 will be a long time around :D

sudoaza:
@ethicalhack3r thanks for the heads up. Checked with the new version and doesnt happen.
Actually i think the issue is with the regex that matches the <link> that eats the next <link> under some strange condition (title attribute has a &raquo; in it) and uses that url which is an ico. But this patch will at least prevent failure and continue the scan.

```ruby
    def rss_url
      homepage_body = Browser.get(@uri.to_s).body
      # Format: <link rel="alternate" type="application/rss+xml" title=".*" href=".*" />
      homepage_body[%r{<link\s*.*\s*type=['|"]application\/rss\+xml['|"]\s*.*\stitle=".*" href=['|"]([^"]+)['|"]\s*\/?>}i, 1]
    end


```

ethicalhack3r:
@FireFart v2 master will be lost though, when we swap to v3, right?

Even if we keep v2 master somewhere, not many people ate

ethicalhack3r:
Accidentally commented and closed. Deleted comment and reopened.

FireFart:
@ethicalhack3r but v2 will live on in the v2 branch so people can still use it. So I would merge it

ethicalhack3r:
Merged, thank you @sudoaza! :)

PR #1218, opened 2018-10-02
Use LABEL and Make use of multi-stage build

mostafahussein:
Maintainer keyword should be replaced with LABEL, and Also using multistage build decreases the image size from 139MB to 117MB

Signed-off-by: Mostafa Hussein <mostafa.hussein91@gmail.com>

## Licensing

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,10 +9,8 @@ RUN echo "gem: --no-ri --no-rdoc" > /etc/gemrc
 COPY . /wpscan
 RUN chown -R wpscan:wpscan /wpscan
 
-# runtime dependencies
-RUN apk add --no-cache libcurl procps sqlite-libs && \
-  # build dependencies
-  apk add --no-cache --virtual build-deps git libcurl ruby-dev libffi-dev make gcc musl-dev zlib-dev procps sqlite-dev && \
+# build dependencies
+RUN apk add --no-cache --virtual build-deps git libcurl ruby-dev libffi-dev make gcc musl-dev zlib-dev procps sqlite-dev && \
   bundle install --system --gemfile=/wpscan/Gemfile $BUNDLER_ARGS && \
   apk del --no-cache build-deps
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,9 +10,8 @@ COPY . /wpscan
 RUN chown -R wpscan:wpscan /wpscan
 
 # build dependencies
-RUN apk add --no-cache --virtual build-deps git libcurl ruby-dev libffi-dev make gcc musl-dev zlib-dev procps sqlite-dev && \
-  bundle install --system --gemfile=/wpscan/Gemfile $BUNDLER_ARGS && \
-  apk del --no-cache build-deps
+RUN apk add --no-cache git libcurl ruby-dev libffi-dev make gcc musl-dev zlib-dev procps sqlite-dev && \
+  bundle install --system --gemfile=/wpscan/Gemfile $BUNDLER_ARGS
 
 WORKDIR /wpscan
 RUN rake install --trace
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ WORKDIR /wpscan
 RUN rake install --trace
 
 FROM ruby:2.5-alpine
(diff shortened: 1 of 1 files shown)


mostafahussein:
Is this a known issue during travis-ci build ?

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated

```

FireFart:
If we use a multi stage build the runtime dependencies and the removal of the build dependencies should not be needed and can be removed (i can not comment on the lines of code here as they were not changed.

Dockerfile
```diff
@@ -19,9 +19,20 @@ RUN apk add --no-cache libcurl procps sqlite-libs && \
 WORKDIR /wpscan
 RUN rake install --trace
 
+FROM ruby:2.5-alpine
+LABEL maintainer="WPScan Team <team@wpscan.org>, Mostafa Hussein <mostafa.hussein91@gmail.com>"
```
FireFart: Please leave the maintainer as WPScan. There is a difference between contributor to the image and maintainer of the image

mostafahussein:
@FireFart  Could you please review it again ?

mostafahussein:
@FireFart  I have removed the runtime dependencies from the build stage. So we have only build dependencies in the build stage and only runtime dependencies in the final stage

Dockerfile
```diff
@@ -9,10 +9,8 @@ RUN echo "gem: --no-ri --no-rdoc" > /etc/gemrc
 COPY . /wpscan
 RUN chown -R wpscan:wpscan /wpscan
 
-# runtime dependencies
-RUN apk add --no-cache libcurl procps sqlite-libs && \
-  # build dependencies
-  apk add --no-cache --virtual build-deps git libcurl ruby-dev libffi-dev make gcc musl-dev zlib-dev procps sqlite-dev && \
+# build dependencies
+RUN apk add --no-cache --virtual build-deps git libcurl ruby-dev libffi-dev make gcc musl-dev zlib-dev procps sqlite-dev && \
```
FireFart: no need for `--virtual`

Dockerfile
```diff
@@ -9,10 +9,8 @@ RUN echo "gem: --no-ri --no-rdoc" > /etc/gemrc
...
   bundle install --system --gemfile=/wpscan/Gemfile $BUNDLER_ARGS && \
   apk del --no-cache build-deps
```
FireFart: this line can be removed as it's only the builder image. no need to cleanup

mostafahussein: thanks for the tip ! done

Dockerfile
mostafahussein: removed, thanks !

FireFart:
Doesn't build on my machine:
```
Step 19/21 : RUN /usr/local/bundle/bin/wpscan --update --verbose
 ---> Running in 24f6d54fa4a7
/usr/local/bundle/bin/wpscan:23:in `load': cannot load such file -- /usr/local/bundle/gems/wpscan-3.3.1/bin/wpscan (LoadError)
	from /usr/local/bundle/bin/wpscan:23:in `<main>'
The command '/bin/sh -c /usr/local/bundle/bin/wpscan --update --verbose' returned a non-zero code: 1
```

Dockerfile
```diff
@@ -9,19 +9,27 @@ RUN echo "gem: --no-ri --no-rdoc" > /etc/gemrc
 COPY . /wpscan
 RUN chown -R wpscan:wpscan /wpscan
```
FireFart: this and the adduser command are not necessary any more in the builder image

FireFart:
It looks like the gems are missing read permissions for users:
```
/ $ ls -alh /usr/local/bundle/gems/wpscan-3.3.1/bin/wpscan 
-rwxrwx--x    1 root     root         402 Oct  2 14:49 /usr/local/bundle/gems/wpscan-3.3.1/bin/wpscan
```

FireFart:
Will fix the error in a follow up. Thanks!

FireFart:
btw here is the fixing commit:
https://github.com/wpscanteam/wpscan/commit/e0fd79f80017d0024b95aafae4a27521ef0f3e49

PR #1233, opened 2018-10-25
Fixed pattern matching on target.wordpress_hosted

lp1dev:
The pattern-matching in the wordpress_hosted attribute was too open, 
for instance the website https://mytest-wordpress.com would be recognized as a wordpress-hosted website, which it isn't.

In addition, the dots were not escaped in the regex, which means https://mytest-wordpress-com would match as well (if it was a correct fqdn).

The regex has been replaced by /\.wordpress\.com$/i  .

## Licensing

### DIFF
--- a/lib/wpscan/target/platform/wordpress.rb
+++ b/lib/wpscan/target/platform/wordpress.rb
@@ -41,7 +41,7 @@ module WPScan
         end
 
         def wordpress_hosted?
-          uri.host =~ /wordpress.com$/i ? true : false
+          uri.host =~ /\.wordpress\.com$/i ? true : false
         end
 
         # @param [ String ] username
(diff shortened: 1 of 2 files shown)


erwanlr:
Thanks for the PR!

Could you also write the associated specs please ?

ethicalhack3r:
Thanks for the PR @lp1dev!

Please can you update the specs so that we can accept your PR?

They can be found here https://github.com/wpscanteam/wpscan/blob/d268a86795f178fa60f0f9f1e53a3fd08de860ba/spec/shared_examples/target/platform/wordpress.rb

lp1dev:
Just to be sure about the best way to test this, is it OK if I add another test under "#wordpress_hosted" where the url is something like 'http://ex-wordpress.com' and wordpress_hosted should be false ?

erwanlr:
Yes :)

ethicalhack3r:
Many thanks @lp1dev! :D

PR #1238, opened 2018-11-01
Typo Intagram -> Instagram

yuliethc:
## Licensing


erwanlr:
This file is taken directly from the author of the plugin, and used for specs. So leaving it as it is is important :)

PR #1367, opened 2019-07-12
Faster Dockerfile build using less layers

felipenazario:
Refactor the Dockerfile to archieve a faster build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,41 +1,80 @@
+# ==============================================================================
+# ================================= BUILD STEP =================================
+# ==============================================================================
 FROM ruby:2.6.2-alpine3.9 AS builder
-LABEL maintainer="WPScan Team <team@wpscan.org>"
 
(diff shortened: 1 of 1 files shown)


NicolasCARPi:
I would put the version in an ENV and use it in the LABEL. This way bumping version only requires one change.

felipenazario:
If a use ENV it means that it will be seen during the runtime of the container and sometime it will be consumed somewhere in the code, but this is not the scenario. So, why keep this at ENV ?

I put it on LABEL tag because it will belongs to the metadata area, which makes more sense to me.

FireFart:
Sorry but I do not like these changes. For example you are creating the user right after installing packages which does not make sense as it is cached in the previous layer. That's the benefit of having layers, we do not need to run everything again and it's cached instead. Also you put the version hardcoded in the dockerfile which makes it harder to release new versions as we also need to update this file. So after all, this change does not really help, sorry. Reducing the layers does not mean gaining speed.

PR #1455, opened 2020-02-17
updater: don't change db if --no-update is passed

jonchang:
If `--no-update` is explicitly passed on the command line, don't attempt to modify the wpscan database.

## Licensing

### DIFF
--- a/lib/wpscan/db/updater.rb
+++ b/lib/wpscan/db/updater.rb
@@ -24,6 +24,7 @@ module WPScan
 
         FileUtils.mkdir_p(repo_directory.to_s) unless Dir.exist?(repo_directory.to_s)
 
+        return if ParsedCli.update == "false"
         raise "#{repo_directory} is not writable" unless repo_directory.writable?
 
         delete_old_files


lib/wpscan/db/updater.rb
```diff
@@ -24,6 +24,7 @@ def initialize(repo_directory)
 
         FileUtils.mkdir_p(repo_directory.to_s) unless Dir.exist?(repo_directory.to_s)
 
+        return if ParsedCli.update == "false"
```
mistydemeo: I think this should be a `false` literal, not a string:
```suggestion
        return if ParsedCli.update == false
```

erwanlr:
Sorry, I can not accept that PR. The directory needs to be writable (especially for the first run of WPScan, or when there is a missing DB file).

Seems like you got that covered via your patch in Homebrew so I will close it.

jonchang:
I'd like to ask you to reconsider.

> Sorry, I can not accept that PR. The directory needs to be writable 

I am aware that this directory needs to be writeable. However, it doesn't make sense that this directory needs to be written to when `--no-update` is explicitly passed to the command line. I would expect that this flag not make any changes to the database.

> (especially for the first run of WPScan, or when there is a missing DB file).

We explicitly update the database for users upon install. This is patch aims to prevent modifications during our build-time tests, which are not permitted to modify the user's environment due to the sandboxed security model.

> Seems like you got that covered via your patch in Homebrew so I will close it.

The patch submitted is the exact patch we use in Homebrew, and we aim to not carry patches around, as they often break when upstream modifies their code. I'd like to point out that until this pull request was submitted, `wpscan` had not been updated since version 2.9.4 (September 2018) due to failing tests.

PR #1757, opened 2022-10-31
Update Daily API Request Limit

Aws0mee:
There are multiple references to an API request limit of 25, but on the website it appears to be 75.
image


erwanlr:
Thx!

PR #1783, opened 2023-05-17
Fixed using proper proxy auth credentials when updating the database and talking to the WPScan API.

ArSn:
The TL;DR of this is that this was calling [a wrong function](https://github.com/wpscanteam/CMSScanner/blob/master/lib/cms_scanner/browser.rb#L42) where the `proxyuserpwd` parameter [does not get added](https://github.com/wpscanteam/CMSScanner/blob/master/lib/cms_scanner/browser.rb#L79) in the CMSScanner gem.

## Licensing


jwidavid:
Love simple fixes!

jwidavid (approved):

PR #1787, opened 2023-06-18
Fixed #1759

### LINKED ISSUE #1759: Database exports not detected
### Subject of the issue
Database exports not detected.

### Your environment
* Version of WPScan: 3.8.22
* Version of Ruby: 2.6.8p205
* Operating System (OS): macOS

### Steps to reproduce
I've created a test environment with some mocked vulnerabilities to test.
Two database export exists in the website, located in the following paths:
- https://wordpress.vulnerable.arall.net/wp-content/uploads/dump.sql
- https://wordpress.vulnerable.arall.net/backups/wordpress.vulnerable.arall.zip

But when I run:
```
wpscan --url https://wordpress.vulnerable.arall.net --enumerate dbe
```
It only reports the last one:

```
[!] https://wordpress.vulnerable.arall.net/backups/wordpress.vulnerable.arall.zip
 | Found By: Direct Access (Aggressive Detection)
```

### Expected behavior
WPScan should report **both** database exports.

### Actual behavior
WPScan **only** reports **one** database export.

### What have you already tried
I monitor the HTTP calls with a proxy, and both files are reached.

What I think the issue is that WPScan is sending a header with `Range: bytes=0-3000`, and expect a `200` response. 
However, the server is responding with `206 Partial Content`.

0n1shi:
#1759

### DIFF
--- a/app/finders/db_exports/known_locations.rb
+++ b/app/finders/db_exports/known_locations.rb
@@ -7,6 +7,10 @@ module WPScan
       class KnownLocations < CMSScanner::Finders::Finder
         include CMSScanner::Finders::Finder::Enumerator
 
+        def valid_response_codes
+          @valid_response_codes ||= [200, 206].freeze
+        end
+
         SQL_PATTERN = /(?:DROP|(?:UN)?LOCK|CREATE|ALTER) (?:TABLE|DATABASE)|INSERT INTO/.freeze
 
         # @param [ Hash ] opts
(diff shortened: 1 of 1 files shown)


app/finders/db_exports/known_locations.rb
```diff
@@ -17,7 +21,7 @@ class KnownLocations < CMSScanner::Finders::Finder
         def aggressive(opts = {})
           found = []
 
-          enumerate(potential_urls(opts), opts.merge(check_full_response: 200)) do |res|
+          enumerate(potential_urls(opts), opts.merge(check_full_response: [200, 206])) do |res|
```
alexsanford, line 24: Could we use `valid_response_codes` here, rather than repeating the array?

alexsanford (commented):
@0n1shi I just tested this, and it works well!

One small change request, and then we should be able to get this merged.

app/finders/db_exports/known_locations.rb
0n1shi: @alexsanford 

Thank you for the review !
I thought I should follow the other finders but using `valid_response_codes` is much better, I will fix it :)

alexsanford: Great, thanks!

alexsanford (approved):
Tested, works fine for me. Will merge once the build finishes.

PR #1839, opened 2024-04-25
Update README.md: Add a helpful tip

redtrib3:
Hi, 
This is an update to the readme.md: docker install section.

I have added a helpful alias command so that we don't have to retype docker command everytime.
This will be useful for the beginner GNU/Linux users.

Feel free to edit anything.

Thanks
- redtrib3

## Licensing

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,14 @@ docker run -it --rm wpscanteam/wpscan --url https://target.tld/ --enumerate u1-1
 
 ** replace u1-100 with a range of your choice.
 
+Instead of running the whole command everytime, we suggest aliasing it to something memorable:
+
+Append the following line to your ~/.bashrc file.
+
+```bash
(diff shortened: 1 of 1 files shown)


akirataguchi115 (commented):
I agree that this is actually convenient for the GNU users. I think though that

- this should not be the responsibility of wpscanteam to tell simple terminal work best practices
- this bloats the valuable readme more than it gives value
- there is a conflict if the user decides to install the ruby version of wpscan on top of the docker version
- this only works if the person uses .bashrc, to give some context, there are other shells like powershell, zsh, windows cmd all of which don't use .bashrc.

Lemme know what you think of my comments. Thanks again for making the pull request!

redtrib3:
Got it, I Agree it's better off to keep it simple. Thanks for the review :)

Issue #1827, opened 2024-01-16
Prompt "Do you want to update now?" not printed


gnuletik:
### Subject of the issue

When running wpscan with an old database, the following message should be printed:

https://github.com/wpscanteam/wpscan/blob/2613fdcc6b22ac7581f3aeb7f33447e3d8417f6c/app/controllers/core.rb#L40

### Your environment

* Version of WPScan: 3.8.25
* Version of Ruby: 3.3.0
* Operating System (OS): Mac OS X 14.2.1
* Terminal Emulator: iTerm2 (Build 3.4.23)

### Steps to reproduce

To reproduce, I set on old date in the `~/.wpscan/db/.last_update` file.

### Expected behavior

The message should have been printed.

It seems to be the same issue as https://github.com/wpscanteam/wpscan/issues/1724 but no solution was found.

### Actual behavior

As you can see, the message is not printed.

```
$ wpscan --url https://localhost
_______________________________________________________________
         __          _______   _____
         \ \        / /  __ \ / ____|
          \ \  /\  / /| |__) | (___   ___  __ _ _ __ ®
           \ \/  \/ / |  ___/ \___ \ / __|/ _` | '_ \
            \  /\  /  | |     ____) | (__| (_| | | | |
             \/  \/   |_|    |_____/ \___|\__,_|_| |_|

         WordPress Security Scanner by the WPScan Team
                         Version 3.8.25
       Sponsored by Automattic - https://automattic.com/
       @_WPScan_, @ethicalhack3r, @erwan_lr, @firefart
_______________________________________________________________

[i] It seems like you have not updated the database for some time.

```

### What have you already tried

* Update WPScan to the latest version [x]
* Update Ruby to the latest version [x]
* Ensure you are using a supported Operating System (Linux and macOS) [x]

alexsanford:
Reproduced on ruby `3.3.7`.

Workaround: run `wpscan --update` manually, then run the desired `wpscan` command.

dkmyta:
@alexsanford Are you still able to reproduce? Although I am still unable to generate the required dir/file to manually modify `.last_update`, If I comment out everything in `/app/controllers/core.rb` in `def update_db_requried?` but the final `output()` and `print` then run `wpscan --url <url>` both lines are output rather than just the first as was originally reported:
```
_______________________________________________________________
         __          _______   _____
         \ \        / /  __ \ / ____|
          \ \  /\  / /| |__) | (___   ___  __ _ _ __ ®
[...]

PR #1905, opened 2025-03-27
Ensure update db now prompt is printed

dkmyta:
## Fixes

Issue: https://github.com/wpscanteam/wpscan/issues/1827

As of ruby `3.3.0`, `update_db_required` no longer prints the last prompt. This is apparently due to behavorial changes introduced around `Readline.readline`.

Initially I attempted to maintain the use of `Readline.readline` but adjust it to account for the the changes, however, I noticed the following warning in CI builds:
```
/opt/hostedtoolcache/Ruby/3.4.2/x64/lib/ruby/3.4.0/readline.rb:4: warning: reline was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add reline to your Gemfile or gemspec to silence this warning.
```
Therefore I've proactively opted to circumnavigate these issues with a secondary approach using `$stdin.gets`.

## Testing
- Checkout this branch
- Ensure you a running ruby `3.3.0`
- Comment out all but the last 3 lines of `update_db_required`
- Run `wpscan --update`
- Verify that all expected content is output and respect the provided request
- Revert to version of ruby prior to `3.3.0` (I've confirmed that version `3.2.2` work without issue previously), and ensure the missing prompt is still output running the same command

### DIFF
--- a/app/controllers/core.rb
+++ b/app/controllers/core.rb
@@ -38,13 +38,11 @@ module WPScan
 
         output('@notice', msg: 'It seems like you have not updated the database for some time.')
         print '[?] Do you want to update now? [Y]es [N]o, default: [N]'
-        # $stdout.flush
+        $stdout.flush
 
-        # response = $stdin.gets.to_s.strip
-        #
-        # !!/^y/i.match?(response)
+        response = $stdin.gets.to_s.strip
 
-        /^y/i.match?(Readline.readline)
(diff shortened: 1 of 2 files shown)


nateweller (approved):
Tested with Ruby `3.2.0` and `3.3.0` with success :+1:

Tests passing with `bundle exec rspec` :+1:

Changes LGTM 🚢

PR #1962, opened 2026-04-20
Don't push daily docker images even though nothing changed.

ArSn:
I also tried to find out why exactly was added, but [the commit adding it](https://github.com/wpscanteam/wpscan/commit/9db97d07301fd90942e3efe59b5737caa51cb1c3) does not really give any context about that, unfortunately.

I have also removed it running on master pushes, I think this was probably a misunderstanding of the "latest" tag, which is supposed to have the latest tagged version in that latest tag, and not always the latest code from master.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,9 +1,6 @@
 name: Build Docker Images
 
 on:
-  push:
-    branches:
-      - master
   release:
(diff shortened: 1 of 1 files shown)


Initsogar (approved):
Thanks, I agree that "latest" should refer to the latest stable release, not just master.

PR #1973, opened 2026-04-22
Updated ruby 3.0 requirements to at least 3.3.

ArSn:
## Proposed changes

* Set rubocop target version to at least 3.3 (this tells rubocop which ruby syntax features are available on all versions, so it needs to stay on the lowest supported version)
* Update gempush ruby version to 4.0.
* Update gemspec to require least 3.3.

* Follow-up to https://github.com/wpscanteam/wpscan/pull/1971 where I missed a few places to adjust.

## Testing instructions

CI

### DIFF
--- a/.github/workflows/gempush.yml
+++ b/.github/workflows/gempush.yml
@@ -11,22 +11,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@master
-    - name: Set up Ruby 3.0
+    - name: Set up Ruby 4.0
       uses: ruby/setup-ruby@v1
       with:
-        ruby-version: '3.0'
-
-    #- name: Publish to GPR
-    #  run: |
-    #    mkdir -p $HOME/.gem
-    #    touch $HOME/.gem/credentials
-    #    chmod 0600 $HOME/.gem/credentials
-    #    printf -- "---\n:github: Bearer ${GEM_HOST_API_KEY}\n" > $HOME/.gem/credentials
-    #    gem build *.gemspec
-    #    gem push --KEY github --host https://rubygems.pkg.github.com/${OWNER} *.gem
-    #  env:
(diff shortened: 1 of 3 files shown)


.github/workflows/gempush.yml
```diff
@@ -11,22 +11,10 @@ jobs:
...
-    #  run: |
-    #    mkdir -p $HOME/.gem
-    #    touch $HOME/.gem/credentials
-    #    chmod 0600 $HOME/.gem/credentials
-    #    printf -- "---\n:github: Bearer ${GEM_HOST_API_KEY}\n" > $HOME/.gem/credentials
-    #    gem build *.gemspec
-    #    gem push --KEY github --host https://rubygems.pkg.github.com/${OWNER} *.gem
-    #  env:
-    #    GEM_HOST_API_KEY: ${{secrets.GITHUB_TOKEN}}
-    #    OWNER: wpscanteam
```
ArSn, line 29: removed this since it was commented out in 2020, so let's not keep the dead code around

wpscan.gemspec
```diff
@@ -21,14 +21,8 @@ Gem::Specification.new do |s|
...
-
-  # Fixes
-  # - warning: ostruct was loaded from the standard library
-  # - warning: fiddle was loaded from the standard library
-  if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('3.3')
-    s.add_dependency('ostruct', '~> 0.6')
-    s.add_dependency('fiddle', '~> 1.1')
-  end
+  s.add_dependency('ostruct', '~> 0.6')
+  s.add_dependency('fiddle', '~> 1.1')
```
ArSn, line 25: since we're always at least 3.3 now, no need for the `if` anymore

Initsogar (approved):

PR #1975, opened 2026-04-22
Added --proxy-target-only CLI option that allows to only send requests to the target through the proxy, and send others (data.wpscan.org and WPScan API) without the proxy directly to the internet.

### LINKED ISSUE #1834: Using proxy to access target without internet
Hello,

I have a web site to scan which is accessible only using a local proxy. I would like to scan all plugins on that web site, so I need to use the api-token I already have. 

If I use the following command, it's doing a basic scan:
`wpscan --url http://target.local --proxy socks5://127.0.0.1:1080`

However, I don't have internet access when using the proxy, so if I want to use `-api-token` parameter, wpscan fails to connect to wpscan.com. 

Is there any way to get wpscan to use the proxy only for accessing the target but not api-token validation?

ArSn:
Closes https://github.com/wpscanteam/wpscan/issues/1834

## Proposed changes

  * Add a new `--proxy-target-only` boolean CLI switch. When used together with `--proxy`, the proxy is applied only to requests made to the target site (`--url`) and is stripped from requests made to the  WPScan API (`wpscan.com`) and the database repository (`data.wpscan.org`). Has no effect unless `--proxy` is also set.
  * Implemented by deleting the `:proxy` and `:proxyuserpwd` keys from the request params in `DB::VulnApi.default_request_params` and `DB::Updater#request_params` when the flag is set.

  The reporter's target site is only reachable via a local proxy (e.g. `socks5://127.0.0.1:1080`), and that proxy has no outbound internet access. Today, setting `--proxy` routes *all* traffic through it, including the WPScan API status/lookup calls and the DB updater hitting `data.wpscan.org`, so the scan aborts before it ever reaches the target. There was no way to scope the proxy to just target traffic.

  `--proxy-target-only` gives users in that topology (air-gapped / split-tunnel / local lab network) a way to use their local proxy for the target while letting API and DB-update requests take the direct internet route.

  ## Testing instructions

  A minimal local reproduction uses [mitmproxy](https://www.mitmproxy.org/) as an allowlist proxy that only lets traffic to the target host through and blocks everything else (so API / DB calls going through it will fail):

  ```bash
  brew install mitmproxy

  # Proxy that only allows requests whose destination host is your target; everything else → 403
  mitmdump -p 8081 --ssl-insecure  --set block_list=':!~d your-target.example.com:403'
  ```

  Then, in a separate shell:

  ```bash
  # 1) Without the new flag: API calls go through the proxy and get blocked.
  #    Expect: scan fails / no vulnerability data, 403s visible in mitmdump output.
  wpscan --url https://your-target.example.com \
         --api-token "$WPSCAN_API_TOKEN" \
         --proxy http://localhost:8081 \
         --disable-tls-checks

  # 2) With the new flag: API calls bypass the proxy and hit wpscan.com directly.
  #    Expect: scan completes, vulnerability data shown, mitmdump only logs target traffic.
  wpscan --url https://your-target.example.com \
         --api-token "$WPSCAN_API_TOKEN" \
         --proxy http://localhost:8081 \
         --proxy-target-only \
         --disable-tls-checks
  ```

  Notes:
  - `--disable-tls-checks` + `--ssl-insecure` are only needed because mitmproxy intercepts TLS; not required in proper proxy setups
  - To exercise the updater path too, force a DB refresh: run `wpscan --update` with and without `--proxy-target-only` — without the flag, the update should fail against the allowlist proxy; with it, the update should succeed.

### DIFF
--- a/app/controllers/vuln_api.rb
+++ b/app/controllers/vuln_api.rb
@@ -11,6 +11,12 @@ module WPScan
           OptString.new(
             ['--api-token TOKEN',
              'The WPScan API Token to display vulnerability data, available at https://wpscan.com/profile']
+          ),
+          OptBoolean.new(
+            ['--proxy-target-only',
+             'When used with --proxy, the proxy is only applied to requests made to the target, ' \
+             'not to requests made to the WPScan API or database repository (data.wpscan.org). ' \
+             'Has no effect unless --proxy is also set.']
           )
         ]
       end
--- a/lib/wpscan/db/updater.rb
+++ b/lib/wpscan/db/updater.rb
@@ -73,13 +73,22 @@ module WPScan
       # @return [ Hash ] The params for Typhoeus::Request
       # @note Those params can't be overriden by CLI options
       def request_params
-        @request_params ||= Browser.instance.default_request_params.merge(
-          timeout: 600,
-          connecttimeout: 300,
-          accept_encoding: 'gzip, deflate',
-          cache_ttl: 0,
-          headers: { 'User-Agent' => Browser.instance.default_user_agent }
-        )
+        @request_params ||= begin
+          params = Browser.instance.default_request_params.merge(
+            timeout: 600,
+            connecttimeout: 300,
+            accept_encoding: 'gzip, deflate',
+            cache_ttl: 0,
+            headers: { 'User-Agent' => Browser.instance.default_user_agent }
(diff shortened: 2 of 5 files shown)


Initsogar (approved):
I was able to reproduce the issue and can confirm that this works as expected!

PR #1976, opened 2026-04-22
Properly set up coveralls parallel build step so we can get per-ruby-version coverage reports and also have proper master builds to compare to.

ArSn:
It was bothering me a little but that each PR says it does not have a coverage report from master to compare to, [example](https://github.com/wpscanteam/wpscan/pull/1975): 

image

so I looked what it would take to fix that. At least one issue seems to be that we run multiple builds and do not have that parallel config set up properly.

There are docs about it here: https://docs.coveralls.io/parallel-builds

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,8 +47,22 @@ jobs:
       if: github.event_name == 'push' && github.ref == 'refs/heads/master'
       run: bundle exec rspec
 
-    - name: Coveralls
-      uses: coverallsapp/github-action@master
+    - name: Coveralls (parallel)
+      uses: coverallsapp/github-action@v2
       continue-on-error: true
       with:
         github-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


ArSn:
> [Coverage Status](https://coveralls.io/builds/79043723)
> 
> Coverage is **85.673%** — **fix-coveralls-setup** into **master**. No base build found for `master`.

to be expected, but if we merge this to master then I think it might work better / compare fine to master in the future 🤞🏻

.github/workflows/build.yml
```diff
@@ -47,8 +47,22 @@ jobs:
       if: github.event_name == 'push' && github.ref == 'refs/heads/master'
       run: bundle exec rspec
 
-    - name: Coveralls
-      uses: coverallsapp/github-action@master
+    - name: Coveralls (parallel)
+      uses: coverallsapp/github-action@v2
```
ArSn, line 51: also changed it to use a proper version rather than just master which might break at any point

Initsogar (approved):

PR #1977, opened 2026-04-23
Merge cms scanner gem into the wpscan project

ArSn:
## Proposed changes

  * Drop the `cms_scanner` gem dependency from `wpscan.gemspec` and fold its transitive runtime deps in directly.
  * Copy all of CMSScanner's `lib/` and `app/` into wpscan under the `WPScan::` namespace (renamed `CMSScanner::` → `WPScan::`, `cms_scanner/` → `wpscan/`, and removed CMSScanner's `NS::` late-binding
  indirection, no longer needed once the framework lives in the same namespace).
  * Collapse ceremonial subclass pairs that existed only because of the gem boundary into single `WPScan::` classes: `Target`, `Browser`, `ParsedCli`, etc. pp.
  * Merge colliding top-level files (`lib/wpscan.rb`, `app/app.rb`, `lib/wpscan/[...].rb`). `lib/wpscan.rb` now owns the framework class  methods (`cached_requests`, `total_requests`, `total_data_sent/received`, `start_memory`) and the `Typhoeus.on_complete` hook that previously lived in `lib/cms_scanner.rb`.
  * Reconcile `.rubocop.yml` and apply `rubocop -A` autocorrects plus targeted hand-fixes.
  * Lift `lib/wpscan/cache/` out of the repo-wide `cache/` `.gitignore` rule so the copied cache handling code files are tracked.

  The `cms_scanner` gem was originally split out so multiple CMS scanners could share a framework, but in practice wpscan is the only consumer. Maintaining the split has real ongoing cost and complexity that we want to get rid of.:

  * **Release coordination**: every fix that crosses the boundary requires two gem releases. 
  * **Ceremonial inheritance**: wpscan touched ~28 distinct `CMSScanner::` symbols, with 9 of them being pure `class WPScan::X < CMSScanner::X` wrappers that added little or nothing. Collapsing those removes a layer of indirection that made the code harder to navigate without buying anything.
  * **No real reusability lost**: wpscan exercises essentially every subsystem of CMSScanner (Target, Browser, Controllers, Finders, Models, OptParseValidator option groups, Scan  orchestration, Vulnerability, References, Error, ExitCode). There was no clean carve-out, keeping the split was preserving form, not function.

  ## Testing instructions

  ```bash
  git checkout merge-cms-scanner
  bundle install                           # should resolve cleanly with no cms_scanner in the set
  bundle exec rspec --tag '~slow'          # expect: 9756 examples, 0 failures, 45 pending
  bundle exec rubocop                      # expect: 341 files inspected, no offenses
  bundle exec rake build                   # runs rubocop + rspec, then builds pkg/wpscan-*.gem
  ```

  Smoke-test the CLI from source:

  ```bash
  bundle exec ruby -Ilib bin/wpscan --help
  bundle exec ruby -Ilib bin/wpscan --version
  bundle exec ruby -Ilib bin/wpscan --url https://<some-wp-site> --no-update --random-user-agent
  ```

  Verify the built gem has no `cms_scanner` dependency:

  ```bash
  gem specification pkg/wpscan-3.8.28.gem | grep "name:"
  # Expect: activesupport, ethon, [...] (plus dev deps). But NO cms_scanner.
  ```


.rubocop.yml
```diff
@@ -15,7 +15,7 @@ Lint/MissingSuper:
 Lint/UriEscapeUnescape:
   Enabled: false
 Metrics/AbcSize:
-  Max: 27
+  Max: 28
```
ArSn, line 18: not exactly great to increase that but we can optimize later on

Initsogar (approved):
Let's give it a try!

PR #1979, opened 2026-04-23
Integrated spec (tests) from the cms_scanner gem so that we have the entire test coverage in the same project.

ArSn:
## Proposed changes

  * Follow-up to https://github.com/wpscanteam/wpscan/pull/1977 to also port over the tests.
  * Port CMSScanner spec tree into wpscan, covering framework code that now lives under `WPScan::`: Browser, Controller, Controllers, Scan, Formatter, Target subcomponents, Finders framework, Cache, models, and app-level formatters/finders/controllers.
  * Reconcile with existing wpscan specs where both repos had coverage (interesting_finding, xml_rpc, vulnerability, typhoeus/response, target).
  * Skip specs for removed mechanisms (`CMSScanner.included` hook, `SubScanner`, `NS::` indirection).
  * Fix a real bug exposed by ported tests: nil guard for `from_vuln_api?` when `effective_url` is nil on timeout responses.

  * Well, we do want to keep the test coverage high, of course 🙂
  * Line coverage: 83.05% → **89.53%** (+6.48pp)
  * Branch coverage: 58.0% → **72.86%** (+14.86pp)

  ## Testing instructions

  Test CI.


Initsogar (approved):
👍🏽

PR #1997, opened 2026-04-28
Improve threshold error messages to mention threshold options

### LINKED ISSUE #1862: Scan Aborted: The number of plugins detected reached the threshold of 100 which might indicate False Positive. It would be recommended to use the --exclude-content-based option to ignore the bad responses.
### Scan aborted due to forced threshold set for plugin detections

WPScan: 3.8.27 (Docker) (Latest)
`docker run -it --rm wpscanteam/wpscan --url (redacted) -e vp --plugins-detection 'aggressive' --api-token (redacted) --exclude-content-based 'not found' --wp-plugins-dir 'app/plugins'`

### Steps to reproduce
I'd exclude my site URL here, but we do indeed have over 100 plugins installed, lots of them are Slider Revolution AddOns that are required to install as dedicated plugins.

### Expected behavior
What's expected is to have the ability to manually override this default behavior

### Actual behavior
image
During Scan, I encountered the error message `Scan Aborted: The number of plugins detected reached the threshold of 100 which might indicate False Positive. It would be recommended to use the --exclude-content-based option to ignore the bad responses`.

### What have you already tried
There's no relevant documentation to override this behavior.

Initsogar:
Fixes https://github.com/wpscanteam/wpscan/issues/1862

## Proposed changes

* Updated `PluginsThresholdReached` error message to mention `--plugins-threshold` option
* Updated `ThemesThresholdReached` error message to mention `--themes-threshold` option
* Both error messages now clearly state that setting the threshold to 0 disables the limit

When users hit the plugins or themes detection threshold (default: 100), they receive an error message that only suggests using `--exclude-content-based` to ignore bad responses. This is misleading because:

1. Users may already be using `--exclude-content-based` (as in the reported issue)
2. The actual solution is to use `--plugins-threshold` or `--themes-threshold` to increase or disable the limit
3. This option exists but is only shown in extended help (`--hh`), making it hard to discover

Sites with legitimately many plugins are currently blocked from scanning without knowing the proper workaround. This change makes the solution discoverable directly in the error message.

## Testing instructions

Since we are only updating documentation, I'd suggest just proofreading the changes.

### DIFF
--- a/lib/wpscan/errors/enumeration.rb
+++ b/lib/wpscan/errors/enumeration.rb
@@ -5,16 +5,16 @@ module WPScan
     class PluginsThresholdReached < Standard
       def to_s
         "The number of plugins detected reached the threshold of #{ParsedCli.plugins_threshold} " \
-          'which might indicate False Positive. It would be recommended to use the --exclude-content-based ' \
-          'option to ignore the bad responses.'
+          'which might indicate False Positive. You can use --plugins-threshold to increase or disable ' \
+          'this limit (set to 0 to disable), or use --exclude-content-based to ignore bad responses.'
       end
     end
 
     class ThemesThresholdReached < Standard
       def to_s
         "The number of themes detected reached the threshold of #{ParsedCli.themes_threshold} " \
-          'which might indicate False Positive. It would be recommended to use the --exclude-content-based ' \
-          'option to ignore the bad responses.'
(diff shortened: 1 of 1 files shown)


ArSn (approved):

SandiyosDev:
Finally, thanks

Issue #1893, opened 2025-02-19
Default cookie jar location is unsafe on a multiuser system (/tmp/ with a predictable name)


hlein:
Before submitting an issue, please make sure you fully read any potential error messages output and did some research on your own.

### Subject of the issue
Use of `/tmp/wpscan/` is predictable and no safe file operations (checking for symlinks, etc.) are done, so `wpscan` is unsafe to run on multiuser systems without using `--cookie-jar` and `--cache-dir`.

### Your environment
* Version of WPScan: HEAD (`20ce10e2ecaeeefc594d8b371de98838c15fa68b`)
* Version of Ruby: `3.1.0`
* Operating System (OS): Gentoo Linux

### Steps to reproduce
Pick a filename in `usera`'s homedir.

As `userb`, create `/tmp/wpscan/` and a symlink to `usera`'s file.

Then as usera, run wpscan.
```
usera@localhost ~ $ ls -ld ~/foo
ls: cannot access '/home/usera/foo': No such file or directory

userb@localhost ~ $ mkdir /tmp/wpscan
userb@localhost ~ $ ln -s /home/usera/foo /tmp/wpscan/cookie_jar.txt

usera@localhost ~ $ wpscan --url https://...

userb@localhost ~ $ ls -l /home/usera/foo
-rw-rw-r-- 1 usera usera 0 Feb 18 17:10 /home/usera/foo
```
This will also clobber/overwrite an existing file like `$HOME/.ssh/authorized_keys`

It's a matter of time to turn this into something significant (use cookie payloads to create interesting contents in `.ssh/authorized_keys` file, etc.). I believe for cookies, `libcurl` does the work - but that library seems to have a (reasonable) expectation that the file paths it is handed are safe.

I think cache filenames are also predictable, if the attacker knows or controls a site that will be visited (filenames based on hashes of URLs). If the attacker controls a site being visited, that could lead to a very easy malicious-file-content attack (serve an invisible `.png` that is actually a replacement `authorized_keys` file, `.bashrc` file, etc.).

### Expected behavior
`wpscan` should either refuse to proceed if the dir cannot be trusted (easier said than done - many TOCTOU concerns), and/or simply use a safer default such as respecting `$TMPDIR` if it is set and prefer that instead of `/tmp/`. On a single-user system or container maybe nobody cares; on a multiuser system the system admin should have already set `$TMPDIR` appropriately.

### Actual behavior
See above.

### What have you already tried
[...]

PR #2003, opened 2026-05-04
fix: Use user cache dir for runtime files

musaabhasan:
Related to #1893

## Proposed changes

* Add a shared `WPScan.user_cache_dir` helper for the per-user XDG cache directory.
* Keep honoring `$TMPDIR` for runtime files when it is explicitly set.
* When `$TMPDIR` is not set, use the per-user XDG cache directory instead of `/tmp/wpscan` for default runtime files such as the HTTP cache and cookie jar.
* Update specs for the default, `$XDG_CACHE_HOME`, and `$TMPDIR` cases.
* Document the runtime file defaults in the README.

* The previous fallback to `/tmp/wpscan` used a predictable shared directory on multi-user systems.
* A user-specific cache directory avoids the default symlink/clobbering risk described in #1893 while preserving explicit `--cache-dir`, `--cookie-jar`, and `$TMPDIR` behavior.

## Testing instructions

* Run `bundle exec rspec spec/lib/controller_spec.rb spec/lib/wpscan_spec.rb`.
* Confirm `--cache-dir` and `--cookie-jar` still override the defaults.
* Confirm the default runtime directory is `$TMPDIR/wpscan` when `$TMPDIR` is set.
* Confirm the default runtime directory is `$XDG_CACHE_HOME/wpscan` or `~/.cache/wpscan` when `$TMPDIR` is not set.

Local validation performed:

* `git diff --check`

I could not run the Ruby test suite locally because Ruby is not installed in this workstation environment.

## Licensing

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,11 @@ The database location follows the [XDG Base Directory Specification](https://spe
 - **New installations**: `~/.cache/wpscan/db` (or `$XDG_CACHE_HOME/wpscan/db` if set)
 - **Existing installations**: `~/.wpscan/db` (legacy path, maintained for backward compatibility)
 
+Runtime files such as the default HTTP cache and cookie jar are stored under `$TMPDIR/wpscan` when
+`$TMPDIR` is set. Otherwise they use the same per-user XDG cache directory, for example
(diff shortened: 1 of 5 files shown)


Initsogar (approved):

Initsogar (approved):
Looks good to me. Thanks for your contribution 🙌

Issue #1688, opened 2021-11-21
Allow for simpler way to skip enumeration


fitzg2:
**Is your feature request related to a problem? Please describe.**
Sometimes we have admin access to a certain amount of WP sites and want to skip enumeration in order to obtain faster results.
Besides faster we also want more precise results. For example we have 12 and we get 6 with or without enumerating. Those other 6 could also be attack vectors. In our example WordFence gets ignored with or without enumeration.

**Describe the solution you'd like**
Not sure. A plugin maybe that precisely lists all plugins and themes plus WP version? the information protected by IP whitelist.
API access to WP?
SSH access?

**Describe alternatives you've considered**
None

**Additional context**
None

erwanlr:
It's quite unclear what you want here, you mention skipping the enumeration but still want to find more plugins. WPScan has a lot of options, the current default being set to provide the best trade between result and speed but you can change any of them to get more results and more accurate one, which will take more time though. `wpscan --hh` will list all available options

If you have admin access, then you could simply install our plugin - https://wordpress.org/plugins/wpscan/

vansh1:
can you tell how to skip enumerate I'm trying to bruteforce but its annoying everytime i have to do enumeration part ehich takes time

fitzg2:
@vansh1 your comment is even more unclear than mine. We have access to our sites as admin and want the scanner to login and get the plugins and themes instead of bruteforcing or scanning. 

By the way we use a wp-scan wrapper called wp-watcher so idk who to talk to.

vansh1:
@fitzg2 sorry dude i came in between of your thread actually mine question is just as simple as i said i want to skip enumeration part while doing bruteforce, is it possible?

alexsanford:
> We have access to our sites as admin and want the scanner to login and get the plugins and themes instead of bruteforcing or scanning.

Flagging this as a feature request.

> i want to skip enumeration part while doing bruteforce, is it possible?

Added a feature request here: https://github.com/wpscanteam/wpscan/issues/1802

PR #1637, opened 2021-04-26
3.9.0

erwanlr:
- [ ] Update version before merging

PR #2008, opened 2026-05-07
Remove default enumerations

### LINKED ISSUE #799: bruteforce ONLY
It would be really neat to ONLY do the bruteforce, and not enumerate everything every time. If all I want to do is bruteforce the wp logins, I don't need wpscan to rescan the whole site again.

### LINKED ISSUE #1628: Change default behaviour
THESE ARE BREAKING CHANGES

**Change 1**

Currently we run a bunch of things when supplying no arguments.

The proposed change would not run anything when no arguments are supplied, and only run enumeration and other things when the additional arguments are supplied.

**Change 2**

Currently the the default enumeration modes are "mixed", with the exception of plugin enumeration, which is "passive".

A lot of users don't read the documentation before using WPScan and miss this.

The proposed change would change `--plugins-detection` default to `mixed`.

**After changes**

- [ ] Update CLI help
- [ ] Update wiki documentation
- [ ] Update Github readme

### LINKED ISSUE #1802: Add an option to skip enumeration
From here: https://github.com/wpscanteam/wpscan/issues/1688#issuecomment-1018087828

For example, the option would allow a user to run a brute force without enumerating plugins.

Initsogar:
Fixes #1802, fixes #1628, fixes #799
Related to #1688

## Proposed changes

This implements @erwanlr's original changes from PR #1637:
* Remove automatic plugin enumeration (`all_plugins: true`) that ran by default
* Remove automatic config backup detection (`config_backups: true`) that ran by default
* Remove default `passive` mode for `--plugins-detection`
* Remove default `mixed` mode for `--plugins-version-detection`
* Simplify `--stealthy` alias by removing `--plugins-version-detection passive`

Additional fixes:
* Fix nil handling in `enum_users_range` method
* Update tests to reflect no default enumeration behavior
* Update documentation to clarify enumeration requires explicit `-e` flag

> [!NOTE]
> This PR only removes **automatic enumeration** defaults (`all_plugins`, `config_backups`), not detection phases (version, theme, interesting findings). The automatic plugin enumeration was the specific complaint in #1802/#1688 - it was slow and unexpected when users just wanted targeted operations like password attacks.

Users have been consistently reporting that WPScan performs unwanted automatic operations (#1802, #1688). When users want to perform a specific task like password brute forcing or user enumeration, WPScan automatically enumerates all plugins and checks for config backups - operations they didn't request and don't need.

The automatic enumeration also wastes time and can consume API requests unnecessarily.

This change gives users explicit control over what gets enumerated. WPScan now only performs operations that users specifically request, making it more predictable and efficient. Users who want enumeration must explicitly request it with `-e ap,cb` or other options.

### Test 1: Verify no default enumeration
```bash
# Run without -e flag
bundle exec ruby -Ilib bin/wpscan --url https://example.com

# Should NOT enumerate plugins or config backups
# Should only do version detection, theme detection, and interesting findings
```

### Test 2: Verify explicit enumeration still works
```bash
# Run with explicit enumeration
bundle exec ruby -Ilib bin/wpscan --url https://example.com -e ap,cb

# Should enumerate all plugins and config backups as requested
```

### Test 3: Verify stealthy mode
```bash
# Run with --stealthy
bundle exec ruby -Ilib bin/wpscan --url https://example.com --stealthy

# Should use random user agent and passive detection mode
# Should NOT set plugins-version-detection (no longer part of stealthy)
```

### Test 4: Verify password attack without enumeration
```bash
# Run password attack only
bundle exec ruby -Ilib bin/wpscan --url https://example.com -U admin -P passwords.txt

# Should proceed directly to password attack without plugin enumeration
# This addresses the original user complaint in #1802
```

## Licensing


ArSn (approved):

Haxel0rd:
Thank you very much for implementing this!

PR #2009, opened 2026-05-07
Add HTTP status code tracking and warning for excessive errors

### LINKED ISSUE #1699: Detect too many 4xx and/or 5xx HTTP error codes
I ran into this issue today where WPScan would not return meaningful results without apparent reasons.

After investigation it was because of two issues:

- The website was losing connection to the database after a lot of requests
- There was some rate limiting / WAF setup at the host provider

Technically the behavior was the following:

- A `500` error when WordPress could not reach the database (`Error establishing a database connection`)
- A `429` when there was too many requests

I was able to workaround thoses isues with a combination of using `--throttle` and setting a custom user-agent using `--ua`.

Ideally wpscan would detect that something unusual is going on and would warn the user.

One way to do that would be to trigger an error message when there are too many 4XX and 5XX error codes like it is done in sqlmap: https://github.com/sqlmapproject/sqlmap/blob/519c0ac01ba0efb82b688ba2381e9e909127985a/lib/core/common.py#L3704

Initsogar:
Fixes #1699

## Proposed changes

* Implement thread-safe HTTP status code tracking via new `HttpStatusTracking` module
* Track all HTTP response codes including failed requests (code 0 for network failures/timeouts) thread-safely
* Display top 5 most common status codes in scan output
* Detect and warn about concerning error patterns with context-specific messages:
  - Rate limiting (429) → "Consider using --throttle or reducing --max-threads"
  - Server errors (5xx) → "Target may be experiencing issues or blocking requests"
  - Failed requests (code 0) → "Could indicate network issues, WAF/IPS blocking, or unavailable target"
  - High error rate → "Could indicate network issues, rate limiting, or security protection"
* Exclude 404 errors from warnings (expected during enumeration)
* Add test coverage

CleanShot 2026-05-07 at 15 27 02@2x

CleanShot 2026-05-07 at 15 43 36@2x

Currently, WPScan fails silently when encountering rate limiting, WAF protection, or server errors, providing no indication to users that their scan results may be incomplete. This leads to confusion and wasted time as users must manually investigate why scans produce no meaningful results.

By tracking and displaying HTTP status codes, users can immediately see:

- If their scan was affected by rate limiting (HTTP 429)
- If a WAF is blocking requests (various 4xx/5xx codes)
- If the server is experiencing issues (HTTP 5xx)
- If requests are failing completely (code 0) due to network issues or aggressive blocking
- The overall distribution of response codes

This helps users understand they need to adjust their scanning approach (e.g., use `--throttle`, change user agent) rather than assuming the scan completed successfully.

The implementation explicitly excludes HTTP 404 errors from triggering warnings, as these are expected and normal during plugin/theme enumeration.

## Testing instructions

Trust the CI?

You can also test by scanning your own sites, add rate limiting, enable a WAF, force specific error responses, or even turn off the server mid-scan to see the new status code tracking and warnings in action.

## Licensing


Initsogar:
Just as a heads-up, I addressed some feedback:

1. **"Code 0" → "failed"**: Now displays `failed: 20` instead of `0: 20` for timeouts/network failures
2. **Multiple warnings**: Changed to return an array so all applicable warnings are shown (e.g., if you have both rate limiting AND server errors, you'll see both)

Also added a bonus improvement - 4xx client errors (403/401) now trigger their own warning, since those often indicate access restrictions.

Output example:

```
[+] Finished: Fri May  8 11:37:28 2026
[+] Requests Done: 1849
[+] Cached Requests: 7
[+] Most response codes received: failed: 1571, 403: 205, 404: 49, 200: 23, 302: 1
[!] Too many failed requests (no response) could indicate network issues, WAF/IPS blocking, or an unavailable target
[!] Too many client errors (4xx). This could indicate access restrictions, authentication issues, or WAF blocking
[+] Data Sent: 0 B
[+] Data Received: 488.437 KB
[+] Memory used: 280.047 MB
[+] Elapsed time: 00:00:52
```

ArSn (approved):
👍🏻

PR #2011, opened 2026-05-07
Made it so that --plugins-list and --themes-list take precedence over their respective -e options for vp/ap/p or vt/at/t.

ArSn:
## Proposed changes

  * Make `--plugins-list` and `--themes-list` first-class triggers for plugin/theme enumeration: passing either option now runs the corresponding enumeration even without a matching `--enumerate` choice.
  * When `--plugins-list` is combined with a colliding `--enumerate` choice (`vp`/`ap`/`p`) - or `--themes-list` with `vt`/`at`/`t` - the list option takes precedence, the colliding enumerate keys are stripped, and a `@notice` is emitted naming the ignored choice(s).
  * Updated the `-e/--enumerate` help text to document the override behaviour.
  * Added nil-safety in `enum_message` and the `vulnerable?` post-filters in `enum_plugins`/`enum_themes` so the new list-only code paths don't dereference a nil `ParsedCli.enumerate`.

  * Previously, `--plugins-list` / `--themes-list` only filtered within an already-enabled enumeration. After the recent removal of default enumerations (#2008), running e.g. `wpscan --url … --plugins-list mylist.txt` silently did nothing, which is counter-intuitive.
  * When users combined both options (e.g. `--enumerate ap --plugins-list mylist.txt`), the enumerate choice was silently dropped from the list source but still affected the on-screen wording and (for `vp`) the vulnerable-only post-filter; so behaviour drifted from what the user asked for, with no indication.
  * Making the list options take precedence and warning on collision matches user expectation: "if I gave you an explicit list, use that list."

  ## Testing instructions

  * Manual sanity checks:
    * `wpscan --url <site> --plugins-list woocommerce --detection-mode mixed` → plugin enumeration runs, no notice, no crash.
    * `wpscan --url <site> --enumerate ap --plugins-list woocommerce` → notice printed (`--plugins-list provided; ignoring colliding --enumerate choice(s): all_plugins`), provided list is used, "All Plugins"
  wording suppressed.
    * `wpscan --url <site> --enumerate vp --plugins-list woocommerce --api-token <token>` → notice printed, list used, no vulnerable-only post-filter applied.
    * `wpscan --url <site> --themes-list twentytwentyfour` → theme enumeration runs without `--enumerate`.
    * `wpscan --help` → `-e/--enumerate` shows the new note about list options overriding vp/ap/p and vt/at/t.
    * `wpscan --url <site> --enumerate vp` (no list) → unchanged behaviour.

### DIFF
--- a/app/controllers/enumeration/cli_options.rb
+++ b/app/controllers/enumeration/cli_options.rb
@@ -15,7 +15,7 @@ module WPScan
         [
           OptMultiChoices.new(
             ['-e', '--enumerate [OPTS]', 'Enumeration Process',
-             'Note: --plugins-list overrides vp/ap/p; --themes-list overrides vt/at/t'
+             'Note: --plugins-list overrides vp/ap/p; --themes-list overrides vt/at/t.'],
             choices: {
               vp: OptBoolean.new(['--vulnerable-plugins']),
               ap: OptBoolean.new(['--all-plugins']),
--- a/app/controllers/enumeration/cli_options.rb
+++ b/app/controllers/enumeration/cli_options.rb
@@ -15,9 +15,7 @@ module WPScan
         [
           OptMultiChoices.new(
             ['-e', '--enumerate [OPTS]', 'Enumeration Process',
-             'Note: vp/ap/p are ignored if --plugins-list is provided, ' \
-             'and vt/at/t are ignored if --themes-list is provided. ' \
-             'A notice is printed when this happens.'],
+             'Note: --plugins-list overrides vp/ap/p; --themes-list overrides vt/at/t'
             choices: {
               vp: OptBoolean.new(['--vulnerable-plugins']),
               ap: OptBoolean.new(['--all-plugins']),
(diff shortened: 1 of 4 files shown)


app/controllers/enumeration/cli_options.rb
```diff
@@ -14,7 +14,10 @@ def cli_options
       def cli_enum_choices
         [
           OptMultiChoices.new(
-            ['-e', '--enumerate [OPTS]', 'Enumeration Process'],
+            ['-e', '--enumerate [OPTS]', 'Enumeration Process',
+             'Note: vp/ap/p are ignored if --plugins-list is provided, ' \
+             'and vt/at/t are ignored if --themes-list is provided. ' \
+             'A notice is printed when this happens.'],
```
Initsogar: Maybe this could be simplified by

```suggestion
             'Note: --plugins-list overrides vp/ap/p; --themes-list overrides vt/at/t'
```

what do you think?

Initsogar (approved):
LGTM. Just left a nit comment, not a blocker.

PR #2013, opened 2026-05-07
Added fetching and display of active install count for detected plugins and themes.

ArSn:
Resolves https://github.com/wpscanteam/wpscan/issues/1866

## Proposed changes

  * Add an `active_installs` lookup for identified plugins and themes via the wordpress.org info API (see [WordPress.org_API codex](https://codex.wordpress.org/WordPress.org_API)).
  * Plugins use `https://api.wordpress.org/plugins/info/1.2/?action=plugin_information&request[slug]=<slug>`; themes use
  `https://api.wordpress.org/themes/info/1.2/?action=theme_information&request[slug]=<slug>&request[fields][active_installs]=1` (themes only return the field when explicitly requested).
  * Render the count in CLI output formatted with thousands separators (e.g. `1,000,000`), and emit it as a raw integer in JSON output under the `active_installs` key.
  * Memoize the result per item so the API is hit at most once per plugin/theme.

  ## Why are these changes being made?

  * Resolves [#1866](https://github.com/wpscanteam/wpscan/issues/1866). When assessing a site, the number of active installs is a useful signal: plugins/themes with low install counts are good candidates for closer review since they tend to have had fewer eyes on them. Today users have to feed the slug list into a separate script; surfacing the count directly in WPScan output removes that step.

  ## Testing instructions

* Run a scan and verify the active installs count shows up for a detected plugin/theme as long as it is a w.org one.

PR #2015, opened 2026-05-08
Added fetching of the last_updated date and time from the WordPress.org API for found plugins and themes.

### LINKED ISSUE #1875: Flag unmaintained and unpopular plugins and themes? Or at least a metric about the popularity and last development?
Since it is often possible to fingerprint most used plugins and themes. Is there an easy way to get a metric on popularity (download count, GitHub stars/forks/issues) and maybe more importantly (last release/update/commit)?

I think this could help to set a higher standard for example by having a (enforceable) policy that does not allow for plugins or themes to be used then they have for example less than 10k downloads and/or are not updates for over a year?

This could be detected externally using wpscan. As well as (in the future) be enforced in WordPress with a policy when installing new plugins and themes.

ArSn:
Resolves #1875

  ## Proposed changes

  * Fetch the `last_updated` date for detected plugins/themes from the public WordPress.org info API (same path PR #2013 added for `active_installs`), and prefer it over the local WPScan DB metadata.
  * Render a human-friendly relative-time hint in the CLI (e.g. `3 months ago`).
  * Tag the source in both CLI and JSON so it is obvious where the value came from (`per WordPress.org` / `"last_updated_source": "WordPress.org"`).
  * Normalize formats per channel: JSON gets a stable ISO-8601 string regardless of source; the CLI uses the WordPress.org-style `YYYY-MM-DD H:MMam/pm GMT` for both sources.
  * Extract the WordPress.org lookup helpers (`wordpress_org_api_url`, `wordpress_org_data`, `fetch_wordpress_org_data`, `active_installs`) into a small `WPScan::Model::WpItem::WordpressOrgData` module so  `WpItem` stays more coherent.

  CLI before (last updated came from the WPScan API, if key available):
  ```
   | Last Updated: 2015-05-16T00:00:00.000Z
   | Active Installs: 1,234
  ```

  CLI after:
  ```
   | Last Updated: 2026-04-14 12:01pm GMT (3 months ago, per WordPress.org)
   | Active Installs: 1,234 (per WordPress.org)
  ```

  JSON additions:
  ```json
  "last_updated": "2026-04-14T12:01:00Z",
  "last_updated_relative": "3 months ago",
  "last_updated_source": "WordPress.org"
  ```

  * Issue #1875 asks for visibility into when a detected plugin/theme was last updated.

  ## Testing instructions

Run some scans and verify the output is as expected, both in CLI output and in JSON.


Initsogar (approved):
Looks good to me :)

> Extract the WordPress.org lookup helpers (wordpress_org_api_url, wordpress_org_data, fetch_wordpress_org_data, active_installs) into a small WPScan::Model::WpItem::WordpressOrgData module so WpItem stays more coherent.

👍🏽

PR #2023, opened 2026-05-11
Added more cases (and checking for them) to the acceptance tests.

### LINKED ISSUE #807: Create a WordPress install which shows off wpscan findings
Currently, not all of wpscan's features are well documented. Some are not documented at all, such as wp-config.php backup disclosure.

Would be cool to create a WordPress install with docker or similar which contains as many of the issues we look for as possible.
- a user with a weak password
- vulnerable timthumb file
- vulnerable plugins
- vulnerable themes
- lots of interesting findings
- ...

This would be useful when demo'ing wpscan features and also maybe in a test environment. Maybe even in an automated test environment, fire up docker ensure findings of previous scan equal findings of new scan.

ArSn:
Resolves https://github.com/wpscanteam/wpscan/issues/807

## Proposed changes

* Add a bunch more things to detect to our acceptance tests.

* To ensure that these things continue to work in the future.

## Testing instructions

* CI covers it, but due to the extracted scripts you can also run it locally now (if you really want to):

```
mkdir wordpress-test && cd wordpress-test
bash ../spec/integration/setup_wp_test_env.sh
cd ..
bash spec/integration/run_scan.sh
```

Note that this still does not cover EVERYTHING that WPScan can detect, but we'd need multiple environments setup to facilitate that, and with this PR I wanted to keep a balance between test runtime / complexity and coverage. We can always add more coverage as we go. 🙂

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,113 +57,13 @@ jobs:
       run: |
         mkdir wordpress-test
         cd wordpress-test
+        bash "${GITHUB_WORKSPACE}/spec/integration/setup_wp_test_env.sh"
 
-        # Configure and start DDEV
-        ddev config --project-type=wordpress --php-version=8.1 --docroot=wordpress
-        ddev start
-
-        # Install WordPress
-        ddev exec wp core download --path=wordpress
-        ddev exec wp core install \
-          --path=wordpress \
-          --url=https://wordpress-test.ddev.site \
-          --title="WPScan Integration Test" \
-          --admin_user=admin \
-          --admin_password=password \
-          --admin_email=test@example.com \
-          --skip-email
-
-        # Install vulnerable plugins and theme for testing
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -57,72 +57,13 @@ jobs:
       run: |
         mkdir wordpress-test
         cd wordpress-test
+        bash "${GITHUB_WORKSPACE}/spec/integration/setup_wp_test_env.sh"
```
Initsogar, line 60: thanks,  it looks cleaner now :)

Initsogar (approved):
Looks good to me.

> Note that this still does not cover EVERYTHING that WPScan can detect, but we'd need multiple environments setup to facilitate that, and with this PR I wanted to keep a balance between test runtime / complexity and coverage. We can always add more coverage as we go. 🙂

👍🏽

PR #2026, opened 2026-05-12
Improved error messages about when database updates via cloudflare are making trouble.

### LINKED ISSUE #1681: Additional Error checking on update
We should implement the following checks in the updater so we can better support with tickets:

# Request ID

Cloudflare sends a header with a unique request id `CF-RAY`. We should provide this id in the error message so we can trace the error in the access logs

# Detect Challenge

Cloudflare will challenge certain IPs. We should detect the challenge and print this out in a nice message. To test this we could create a firewall rule to challenge a specific ip

# Detect Blocks

Some IPs are blocked. This results in a cloudflare 1000 error I think. We should detect this and print it out

# Check checksum file

We should do a format check on the hashsum. If it does not match a certain length and character set (or contains multiple lines), there is an http error which we can print out

ArSn:
Fixes https://github.com/wpscanteam/wpscan/issues/1681 - although not the entire WPScan.com API is servied via Cloudflare anymore (which this issue was originally about), we still do it for the database data served from data.wpscan.org; so it's a good idea to improve the error messaging here.

## Proposed changes

  * `Error::Download` and `Error::ChecksumsMismatch` now point users to the status page, contact form, and GitHub issues when the failure is against `data.wpscan.org`.
  * Both errors surface the Cloudflare Ray ID (`CF-Ray` header) when present.
  * Added focused specs under `spec/lib/errors/`.

  * Partially addresses [#1681](https://github.com/wpscanteam/wpscan/issues/1681): the API no longer sits behind Cloudflare, but `data.wpscan.org` does, so DB-update failures are the remaining case worth improving.
  * The Ray ID lets anyone with Cloudflare account access correlate a user report to the exact edge log entry.

  ## Testing instructions

You can change what is returned from `WPScan::DB::Updater#local_file_checksum` or `WPScan::DB::Updater#remote_file_url` to reliably cause errors (then run `wpscan --update`).

### DIFF
--- a/lib/wpscan/db/updater.rb
+++ b/lib/wpscan/db/updater.rb
@@ -136,7 +136,8 @@ module WPScan
         FileUtils.rm(backup_file_path(filename))
       end
 
-      # @return [ String ] The checksum of the downloaded file
+      # @return [ Array(String, String) ] The checksum of the downloaded file and the
+      #                                    Cloudflare Ray ID of the response (or nil)
       def download(filename)
         file_path = local_file_path(filename)
         file_url  = remote_file_url(filename)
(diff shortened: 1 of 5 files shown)


Initsogar (approved):
Works like a charm

PR #2027, opened 2026-05-12
Stream output as we find things for plugins, themes, and users.

### LINKED ISSUE #952: Output enumeration findings as they are found
Currently, when enumeration (plugins, themes, etc) has completed we output the findings. This method is fine when only enumerating themes, or vulnerable plugins, as they don't take much time. However, when enumerating all plugins, this is more of an issue as it takes much longer. For example, there is more chance that someone manually blocks you if they notice you. This would then output nothing and your time so far would have been wasted. Unlikely, but there's also an increased chance (due to the amount of time) that your internet connection fails, power outage or whatever. Also, the tester can begin looking at findings as they come through, rather than having to wait until everything has been enumerated.

I think it would be better to output findings as we find them.

ArSn:
Resolves https://github.com/wpscanteam/wpscan/issues/952

## Proposed changes

  - Enumeration findings (plugins, themes, users) now stream to stdout as each one is discovered, instead of being batched until the enumeration step finishes.
  - Added a `Formatter::Base#streams?` capability flag: `cli`, `cli_no_color` and `jsonl` opt in; `json` and `sarif` keep their batched single-document contract.
  - Split the per-item rendering out of `plugins.erb`/`themes.erb`/`users.erb` into singular `plugin.erb`/`theme.erb`/`user.erb` partials. Plural views delegate to them, so streamed and batched output stay
  byte-identical per item.
  - The CLI singular views emit an ANSI clear-line sequence (TTY only) before each finding so the ruby-progressbar line gets wiped instead of getting stuck in the middle of finding output.

  - A long `--enumerate ap` run can take many minutes and previously produced no output until the very end. If the operator was rate-limited, lost connectivity, or hit Ctrl-C, all results so far were lost and the time was wasted.
  - `--no-stream` exists as an escape hatch for scripts that grep for the existing `[i] Plugin(s) Identified:` headline shape, or anyone who simply prefers the all-at-end format.

## Testing instructions

- Run an enumeration against a site that finds multiple things, for example with `-e p`.
- Verify by default it does stream on the CLI output (which is also default, so no extra parameters needed).
- Verify that passing `--no-stream` still has the previous behavior of not streaming and printing everything once the enumeration is done.
- Verify that it streams for `-f cli-no-color` and `-f jsonl` as well
- Verify that it does NOT stream for `-f json` and `-f sarif`

### DIFF
--- a/app/controllers/core/cli_options.rb
+++ b/app/controllers/core/cli_options.rb
@@ -16,6 +16,11 @@ module WPScan
           OptFilePath.new(['-o', '--output FILE', 'Output to FILE'], writable: true, exists: false),
           OptChoice.new(['-f', '--format FORMAT',
                          'Output results in the format supplied'], choices: formats),
+          OptBoolean.new(['--[no-]stream',
+                          'Emit enumeration findings (plugins/themes/users) as they are discovered, ' \
+                          'instead of waiting until each enumeration step completes. ' \
+                          'Has no effect on the json or sarif output formats, which always batch.'],
+                         default: true),
           OptChoice.new(['--detection-mode MODE'],
                         choices: %w[mixed passive aggressive],
                         normalize: :to_sym,
--- a/app/controllers/enumeration/enum_methods.rb
+++ b/app/controllers/enumeration/enum_methods.rb
@@ -4,6 +4,14 @@ module WPScan
   module Controller
     # Enumeration Methods
     class Enumeration < WPScan::Controller::Base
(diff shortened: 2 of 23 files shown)


Initsogar (approved):
👍🏽

Confirmed:
* Works for the default CLI output, `-f cli-no-color`, and `-f jsonl`.
* Passing `--no-stream` disables streaming.
* Streaming is not supported for `-f json` or `-f sarif`.

PR #2028, opened 2026-05-12
Add --max-retries option for automatic retry of failed brute force requests

### LINKED ISSUE #800: bruteforce retries
sometimes to evade mod_security, I use a bunch of different proxies and I've configured a load balancer to rotate them. Sometimes the proxies barf, or wpscan tries to hit a proxy during the rotation and gets an unexpected result back from the upstream connection. It would be pretty awesome to have a 'retries' functionality available so that it can validate whether the attempt actually succeeded or actually failed by trying for either a valid positive or valid negative response from the wordpress install, rather than a proxy error or a timeout.

Initsogar:
Fixes #800

## Proposed changes

* Added new `--max-retries N` CLI option to automatically retry failed brute force requests
* Modified `BreadthFirstDictionaryAttack#attack` to implement retry logic with configurable maximum attempts
* Added `queue_login_request` helper method to handle retry logic and progress tracking
* Updated progress bar to show retry attempts when `show_progression` is enabled

CleanShot 2026-05-12 at 15 50 43@2x

When performing brute force attacks through proxies or unstable networks, transient failures (timeouts, proxy errors, connection failures) cause passwords to be skipped without being properly tested. 

This feature addresses a long-standing user request (#800, opened in 2015) by providing automatic retry capability for failed requests. The implementation distinguishes between actual authentication failures and network/proxy errors, only retrying the latter. This ensures that valid credentials aren't missed due to transient network issues while maintaining backward compatibility (default: 0 retries).

The feature integrates seamlessly with the existing `--wordlist-skip` option, allowing users to both resume interrupted attacks and handle transient failures automatically.

## Testing instructions

This is a tricky one to test since we need an unstable environment or forcing it by stopping/restarting your web server in the middle of a long-running scan. Or you can just trust the CI.

```
bundle exec ruby -Ilib bin/wpscan --url https://example.com --passwords wordlist.txt --usernames admin --max-retries 3
```

## Licensing

### DIFF
--- a/lib/wpscan/finders/finder/breadth_first_dictionary_attack.rb
+++ b/lib/wpscan/finders/finder/breadth_first_dictionary_attack.rb
@@ -10,6 +10,7 @@ module WPScan
         # @param [ Hash ] opts
         # @option opts [ Boolean ] :show_progression
         # @option opts [ Integer ] :wordlist_skip Number of passwords to skip from the beginning
+        # @option opts [ Integer ] :max_retries Maximum retry attempts for failed requests
         #
         # @yield [ WPScan::User ] When a valid combination is found
         #
@@ -22,6 +23,7 @@ module WPScan
         def attack(users, wordlist_path, opts = {})
           wordlist = File.open(wordlist_path)
           skip_count = opts[:wordlist_skip] || 0
+          max_retries = opts[:max_retries] || 0
 
           # Calculate total, accounting for skipped passwords
           total_passwords = wordlist.count
@@ -33,6 +35,8 @@ module WPScan
           # Keep the number of requests sent for each users
           # to be able to correctly update the progress when a password is found
           user_requests_count = {}
+          # Track retry attempts for each user/password combination
+          retry_count = Hash.new { |h, k| h[k] = Hash.new(0) }
 
           users.each { |u| user_requests_count[u.username] = 0 }
 
@@ -47,30 +51,9 @@ module WPScan
             break if remaining_users.empty?
 
             remaining_users.each do |user|
-              request = login_request(user.username, password)
-
-              user_requests_count[user.username] += 1
-
-              request.on_complete do |res|
-                progress_bar.title = "Trying #{user.username} / #{password}"
-
-                progress_bar.increment unless progress_bar.progress == progress_bar.total
-
-                if valid_credentials?(res)
-                  user.password = password
-
-                  begin
-                    progress_bar.total -= effective_passwords - user_requests_count[user.username]
-                  rescue ProgressBar::InvalidProgressError
-                  end
-
-                  yield user
-                elsif errored_response?(res)
-                  output_error(res)
-                end
+              queue_login_request(user, password, user_requests_count, retry_count, max_retries, effective_passwords, opts) do |found_user|
+                yield found_user if block_given?
               end
-
-              hydra.queue(request)
               queue_count += 1
 
               if queue_count >= hydra.max_concurrency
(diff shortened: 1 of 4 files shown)


lib/wpscan/finders/finder/breadth_first_dictionary_attack.rb
```diff
@@ -47,41 +51,81 @@ def attack(users, wordlist_path, opts = {})
...
+        #
+        # @param [ WPScan::Model::User ] user
+        # @param [ String ] password
+        # @param [ Hash ] user_requests_count
+        # @param [ Hash ] retry_count
+        # @param [ Integer ] max_retries
+        # @param [ Integer ] effective_passwords Total passwords accounting for skip
+        # @param [ Hash ] opts
+        #
+        # rubocop:disable all
```
ArSn: hmm :) might be a good indicator to refactor stuff, maybe

lib/wpscan/finders/finder/breadth_first_dictionary_attack.rb
```diff
@@ -47,41 +51,81 @@ def attack(users, wordlist_path, opts = {})
...
-                if valid_credentials?(res)
-                  user.password = password
+          user_requests_count[user.username] += 1
 
-                  begin
-                    progress_bar.total -= effective_passwords - user_requests_count[user.username]
-                  rescue ProgressBar::InvalidProgressError
-                  end
+          request.on_complete do |res|
+            retry_key = "#{user.username}:#{password}"
```
ArSn: doesn't look used


```suggestion
```

lib/wpscan/finders/finder/breadth_first_dictionary_attack.rb
ArSn: hmm :) a good indicator to refactor this, maybe?

lib/wpscan/finders/finder/breadth_first_dictionary_attack.rb
```diff
@@ -47,41 +51,81 @@ def attack(users, wordlist_path, opts = {})
...
               end
 
-              hydra.queue(request)
-              queue_count += 1
+              yield user if block_given?
+            elsif errored_response?(res)
+              # Error response - potentially retryable
+              if current_retry < max_retries
+                # Retry the request
+                retry_count[user.username][password] += 1
```
ArSn: this should probably be cleared again once we are through with this user, because otherwise it just keeps growing if we have many retries (up to the size of the passwordlist)

Initsogar:
Based on the feedback above, I refactored the password attack implementation (https://github.com/wpscanteam/wpscan/pull/2028/commits/1ff2216f807fecee7bc3a04daf107650a93142d6):
- Extracted `LoginAttempt` class to handle individual login attempts with retry logic
- Extracted `ProgressTracker` class to manage progress bar updates and request counting
- Introduced `LoginConfig` struct to bundle dependencies and reduce parameter passing
- Removed the 8-parameter `queue_login_request` method in favor of clean classes

lib/wpscan/finders/finder/breadth_first_dictionary_attack.rb
Initsogar: Good call. Addressed in https://github.com/wpscanteam/wpscan/commit/1ff2216f807fecee7bc3a04daf107650a93142d6

lib/wpscan/finders/finder/breadth_first_dictionary_attack.rb
Initsogar: Addressed in https://github.com/wpscanteam/wpscan/pull/2028/commits/2b47631ef0a929f4bb2a3fefaea67b5d20534fca

lib/wpscan/finders/finder/breadth_first_dictionary_attack.rb
Initsogar: ty, addressed in https://github.com/wpscanteam/wpscan/pull/2028/commits/bd2df5c18a39eab1d12612caeae2e962827e6bc6

lib/wpscan/finders/finder/breadth_first_dictionary_attack.rb
```diff
@@ -5,11 +5,137 @@ module Finders
...
+            @user = user
+            @password = password
+            @max_retries = max_retries
+            @retry_count = 0
+            @config = config
+          end
+
+          def execute(&block)
+            request = @config.request_builder.call(@user.username, @password)
+            @config.tracker.record_request(@user.username)
```
ArSn:
```suggestion
            @config.tracker.record_request(@user.username) if @retry_count.zero?
```

I think without this we are counting each attempt (instead of each user/password combo), which then in combination with `adjust_total_on_success` might lead to some under- and/or overcounting issues.

The flipside is with this change we dont see multiple attempts in the progress bar, but I think thats fine?

lib/wpscan/finders/finder/breadth_first_dictionary_attack.rb
```diff
@@ -5,11 +5,137 @@ module Finders
...
+            if should_retry?
+              retry_attempt(&)
+            else
+              finalize_failed_attempt(response)
+            end
+          end
+
+          def should_retry?
+            @retry_count < @max_retries && @user.password.nil?
+          end
```
ArSn, line 116:
```suggestion
          def handle_error(response, &)
            if @user.password
              @config.tracker.increment
              return
            end
          
            if should_retry?
              retry_attempt(&)
            else
              finalize_failed_attempt(response)
            end
          end

          def should_retry?
            @retry_count < @max_retries
          end
```

suggesting this change because without it the user will still see the errors before the user then actually was attempted correctly (without error) ... if I got that right, that is 🙂

lib/wpscan/finders/finder/breadth_first_dictionary_attack.rb
Initsogar: Yep, it's fine. Addressed in https://github.com/wpscanteam/wpscan/pull/2028/commits/c66cefa3539ae4641741af2f3aec497ea717f6c7. Thank you 🙌

lib/wpscan/finders/finder/breadth_first_dictionary_attack.rb
Initsogar: Addressed in https://github.com/wpscanteam/wpscan/pull/2028/commits/a06fb5890924b914ee8f97e025dbf4c774e8a82f :)!

ArSn (approved):
All looking good, also verified that I can still successfully find accounts:

```
[!] Valid Combinations Found:
 | Username: testadmin, Password: password4
```

PR #1815, opened 2023-11-12
XDG Support for wpscan db

alichtman:
Fix #1586

This replaces #1780 -- I messed up the rebase and ended up cherrypicking the commit instead of fighting with git / gh :upside_down_face: 

## Licensing

By submitting code contributions to the WPScan development team via Github Pull Requests, or any other method, it is understood that the contributor is offering the WPScan company (company number 	83421476900012), which is registered in France, the unlimited, non-exclusive right to reuse, modify, and relicense the code.

Issue #1586, opened 2020-12-30
XDG Support


MarcinWieczorek:
Hello,
Thank you for maintaining such a wonderful project.
I noticed wpscan creates a directory in $HOME, which creates clutter. There is a standard for that: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
I think that `XDG_DATA_HOME` would be good as an alternative (in most cases that would be `~/.local/share/wpscan/db`.
Some migration and backward compatibility would be required, but I hope someone finds time for that.

The issue seems to be here:
https://github.com/wpscanteam/wpscan/blob/f61c55b350a426dd336ae2cfc9607a0f9e76ecea/lib/wpscan.rb#L40

Thanks!

noraj:
Nice `$HOME/.wpscan/db` was moved to `XDG_CACHE_HOME/wpscan/db` with #1815

However, what #1586 suggested, it to move away the whole `$HOME/.wpscan`. Because even if the db moved to a compliant path, `~/.wpscan/scan.yml` is still in home instead of `XDG_CONFIG_HOME/wpscan/scan.yml` (~/.config/wpscan/scan.yml). What about making wpscan 100% XDG compliant with adding `XDG_CONFIG_HOME/wpscan` to the list of 

https://github.com/wpscanteam/wpscan/blob/1845814cf716332b4a70cde9f167d3455fe1e0db/README.md?plain=1#L155-L158

Update: I tried a PR #2029

PR #2029, opened 2026-05-13
XDG support for config base directory

noraj:
Related to https://github.com/wpscanteam/wpscan/issues/1586#issuecomment-4441989010

XDG Support asked in #1586 was partially supported in #1815 (only database in cache folder)

This PR also brings XDG Support for the config directory

## Proposed changes

Follows https://specifications.freedesktop.org/basedir/latest/ for config

Complete #1815 for #1586 100% support

## Testing instructions

- Create $XDG_CONFIG_HOME env. var. to a custom location and create a scan.{json,yml} there
- Create ~/.config/wpscan/scan.{json,yml}

See if it is taken into account.

Also check previously supported path are still working.

If the PR is accepted, please complete https://github.com/wpscanteam/wpscan/blob/master/spec/lib/controllers_spec.rb#L139-L154

## Licensing

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -196,7 +196,9 @@ WebMock adapter has custom port normalization for Typhoeus (spec/spec_helper.rb:
 
 **Configuration Files:**
 WPScan loads options from (in order):
-1. `~/.wpscan/scan.json` or `~/.wpscan/scan.yml`
-2. `pwd/.wpscan/scan.json` or `pwd/.wpscan/scan.yml`
+1. `$XDG_CONFIG_HOME/wpscan/scan.json` or `$XDG_CONFIG_HOME/wpscan/scan.yml`
+2. `~/.config/wpscan/scan.json` or `~/.config/wpscan/scan.yml`
+3. `~/.wpscan/scan.json` or `~/.wpscan/scan.yml`
+4. `pwd/.wpscan/scan.json` or `pwd/.wpscan/scan.yml`
 
 Use snake_case for CLI options in config (e.g., `api_token`, `max_threads`).
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -196,8 +196,8 @@ WebMock adapter has custom port normalization for Typhoeus (spec/spec_helper.rb:
 
 **Configuration Files:**
 WPScan loads options from (in order):
-1. `$XDG_CONFIG_HOME/wpscan/scan.json` or `$XDG_CONFIG_HOME/wpscan/scan.yml`
-2. `~/.config/wpscan/scan.json` or `~/.config/wpscan/scan.yml`
+1. `$XDG_CONFIG_HOME/wpscan/scan.json` or `$XDG_CONFIG_HOME/wpscan/scan.yml` (if `XDG_CONFIG_HOME` is set)
+2. `~/.config/wpscan/scan.json` or `~/.config/wpscan/scan.yml` (if `XDG_CONFIG_HOME` is not set)
 3. `~/.wpscan/scan.json` or `~/.wpscan/scan.yml`
 4. `pwd/.wpscan/scan.json` or `pwd/.wpscan/scan.yml`
 
--- a/README.md
+++ b/README.md
@@ -152,10 +152,10 @@ Up to **25** API requests per day are given free of charge, that should be suita
 
 WPScan can load all options (including the `--url`) from configuration files, the following locations are checked (order: first to last):
 
-- `$XDG_CONFIG_HOME/wpscan/scan.json`
-- `$XDG_CONFIG_HOME/wpscan/scan.yml`
-- `~/.config/wpscan/scan.json`
-- `~/.config/wpscan/scan.yml`
+- `$XDG_CONFIG_HOME/wpscan/scan.json` (if `XDG_CONFIG_HOME` is set)
+- `$XDG_CONFIG_HOME/wpscan/scan.yml` (if `XDG_CONFIG_HOME` is set)
+- `~/.config/wpscan/scan.json` (if `XDG_CONFIG_HOME` is not set)
+- `~/.config/wpscan/scan.yml` (if `XDG_CONFIG_HOME` is not set)
 - `~/.wpscan/scan.json`
 - `~/.wpscan/scan.yml`
 - `pwd/.wpscan/scan.json`
--- a/README.md
+++ b/README.md
@@ -152,6 +152,10 @@ Up to **25** API requests per day are given free of charge, that should be suita
 
 WPScan can load all options (including the `--url`) from configuration files, the following locations are checked (order: first to last):
 
+- `$XDG_CONFIG_HOME/wpscan/scan.json`
+- `$XDG_CONFIG_HOME/wpscan/scan.yml`
+- `~/.config/wpscan/scan.json`
+- `~/.config/wpscan/scan.yml`
 - `~/.wpscan/scan.json`
 - `~/.wpscan/scan.yml`
 - `pwd/.wpscan/scan.json`
@@ -161,7 +165,7 @@ If those files exist, options from the `cli_options` key will be loaded and over
 
 e.g:
 
-`~/.wpscan/scan.yml`:
+`~/.config/wpscan/scan.yml`:
 
 ```yml
 cli_options:
@@ -189,7 +193,7 @@ cli_options:
 
 ## Save API Token in a file
 
-The feature mentioned above is useful to keep the API Token in a config file and not have to supply it via the CLI each time. To do so, create the ~/.wpscan/scan.yml file containing the below:
+The feature mentioned above is useful to keep the API Token in a config file and not have to supply it via the CLI each time. To do so, create the ~/.config/wpscan/scan.yml file containing the below:
 
 ```yml
 cli_options:
--- a/lib/wpscan/controllers.rb
+++ b/lib/wpscan/controllers.rb
@@ -20,7 +20,7 @@ module WPScan
       ## XDG path
       [ENV['XDG_CONFIG_HOME', Pathname.new(Dir.home).join('.config')], ].each do |dir|
         option_parser.config_files.class.supported_extensions.each do |ext|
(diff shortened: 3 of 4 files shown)


noraj:
Note: [CI is failing](https://github.com/wpscanteam/wpscan/actions/runs/25807212088/job/75813225189?pr=2029) because of the rubocop linter. I'm open to ideas to improve the syntax so `Metrics/AbcSize` cop would be respected without splitting the `register_config_files` method.

lib/wpscan/controllers.rb
ArSn:
```suggestion
      xdg = ENV.fetch('XDG_CONFIG_HOME', nil)
      xdg = Pathname.new(Dir.home).join('.config') if xdg.nil? || xdg.empty?
      app = WPScan.app_name

      dirs = [[xdg, app], [Dir.home, ".#{app}"], [Dir.pwd, ".#{app}"]]
      exts = option_parser.config_files.class.supported_extensions

      dirs.product(exts).each do |(dir, sub), ext|
        option_parser.config_files << Pathname.new(dir).join(sub, "scan.#{ext}").to_s
      end
```

should make it work and also satisfy rubocop?

Note that there also is a runtime bug in the current version: `ENV['XDG_CONFIG_HOME', Pathname.new(Dir.home).join('.config')]` is not valid: `ENV.[]` [takes one argument](https://docs.ruby-lang.org/en/master/ENV.html#method-c-5B-5D) and will raise ArgumentError on first invocation.

ArSn (changes requested):

ArSn:
I'm sorry my code suggestion seems to be scrambled a bit, what I meant so say is that the method should probably look somewhat like this:

```ruby
  def register_config_files
    xdg = ENV.fetch('XDG_CONFIG_HOME', nil)
    xdg = Pathname.new(Dir.home).join('.config') if xdg.nil? || xdg.empty?
    app = WPScan.app_name

    dirs = [[xdg, app], [Dir.home, ".#{app}"], [Dir.pwd, ".#{app}"]]
    exts = option_parser.config_files.class.supported_extensions

    dirs.product(exts).each do |(dir, sub), ext|
      option_parser.config_files << Pathname.new(dir).join(sub, "scan.#{ext}").to_s
    end
  end
```

lib/wpscan/controllers.rb
noraj: > Note that there also is a runtime bug in the current version: `ENV['XDG_CONFIG_HOME', Pathname.new(Dir.home).join('.config')]` is not valid: `ENV.[]` [takes one argument](https://docs.ruby-lang.org/en/master/ENV.html#method-c-5B-5D) and will raise ArgumentError on first invocation.

Good catch, I pasted the second argument at the wrong place. Fixed: https://github.com/wpscanteam/wpscan/pull/2029/changes/cf7e28026bbac0e575ce2ed45bf64c274dcddbf3

noraj: About the conditional assignment

```
      xdg = ENV.fetch('XDG_CONFIG_HOME', nil)
      xdg = Pathname.new(Dir.home).join('.config') if xdg.nil? || xdg.empty?
```

I thought about this at first

```ruby
xdg_path = ENV['XDG_CONFIG_HOME'] || Pathname.new(Dir.home).join('.config')
```

But since the goal of `register_config_files` is to register all possible paths, I wanted to register both of them, not one or the other depending on the env. var. declaration. But now I understand the issue is when the env. var. is not declared this will be an issue because it will store the evaluated value (nil) and not the dynamic env. var..

so we would need something like

```
xdg = ENV.fetch('XDG_CONFIG_HOME', nil)
dirs << xdg unless xdg.nil? || xdg.empty?
dirs << Pathname.new(Dir.home).join('.config')
```

That's the env. var. that should be added only if existing, not the fallback option.

noraj:
> I'm sorry my code suggestion seems to be scrambled a bit, what I meant so say is that the method should probably look somewhat like this:
> 
> ```ruby
>   def register_config_files
>     xdg = ENV.fetch('XDG_CONFIG_HOME', nil)
>     xdg = Pathname.new(Dir.home).join('.config') if xdg.nil? || xdg.empty?
>     app = WPScan.app_name
> 
>     dirs = [[xdg, app], [Dir.home, ".#{app}"], [Dir.pwd, ".#{app}"]]
>     exts = option_parser.config_files.class.supported_extensions
> 
>     dirs.product(exts).each do |(dir, sub), ext|
>       option_parser.config_files << Pathname.new(dir).join(sub, "scan.#{ext}").to_s
>     end
>   end
> ```

Regarding https://github.com/wpscanteam/wpscan/pull/2029#discussion_r3235655291

Maybe something like:

```ruby
  def register_config_files
    xdg_env = ENV.fetch('XDG_CONFIG_HOME', nil)
    xdg_fallback = Pathname.new(Dir.home).join('.config') 
    app = WPScan.app_name

    dirs = [[xdg_fallback, app], [Dir.home, ".#{app}"], [Dir.pwd, ".#{app}"]]
    dirs.unshift([xdg_env, app]) unless xdg.nil? || xdg.empty?
    exts = option_parser.config_files.class.supported_extensions

    dirs.product(exts).each do |(dir, sub), ext|
      option_parser.config_files << Pathname.new(dir).join(sub, "scan.#{ext}").to_s
    end
  end
```

lib/wpscan/controllers.rb
ArSn: I'm afraid this is still not correct.

This iterates over both entries instead of picking one as a fallback, so:

- When XDG_CONFIG_HOME is set → it registers the XDG path **and** ~/.config/wpscan/...
- When XDG_CONFIG_HOME is unset → ENV[...] returns nil, then Pathname.new(nil).join(...) raises TypeError: no implicit conversion of nil into String.

What should happen instead is making the XDG things the default and otherwise fallback to the previous values if that is not configured, as in my suggested code.

And about the rubocop ABC thing - there is more to read here https://docs.rubocop.org/rubocop/latest/cops_metrics.html with some additional references too, but the TL;DR is that there is just too much stuff going on at once, i.e. loops in loops and then calling a different methods in there.

You can always run rubocop locally btw: `bundle exec rubocop` (it runy pretty quickly for the entire project)

ArSn (changes requested):

ArSn:
> Regarding [#2029 (comment)](https://github.com/wpscanteam/wpscan/pull/2029#discussion_r3235655291)
> 
> Maybe something like:
> 
> ```ruby
>   def register_config_files
>     xdg_env = ENV.fetch('XDG_CONFIG_HOME', nil)
>     xdg_fallback = Pathname.new(Dir.home).join('.config') 
>     app = WPScan.app_name
> 
>     dirs = [[xdg_fallback, app], [Dir.home, ".#{app}"], [Dir.pwd, ".#{app}"]]
>     dirs.unshift([xdg_env, app]) unless xdg.nil? || xdg.empty?
>     exts = option_parser.config_files.class.supported_extensions
> 
>     dirs.product(exts).each do |(dir, sub), ext|
>       option_parser.config_files << Pathname.new(dir).join(sub, "scan.#{ext}").to_s
>     end
>   end
> ```


That sounds more like it but you'll have to update the variable names everywhere, for example `unless xdg.nil? || xdg.empty?` needs to be updated too then.

And again like I said in my other comment, we do NOT want to register multiple paths at once. This would potentially mix data from different dirs instead of dealing only with one specific one, leading to more potential trouble.

AGENTS.md
```diff
@@ -196,7 +196,9 @@ WebMock adapter has custom port normalization for Typhoeus (spec/spec_helper.rb:
 
 **Configuration Files:**
 WPScan loads options from (in order):
-1. `~/.wpscan/scan.json` or `~/.wpscan/scan.yml`
-2. `pwd/.wpscan/scan.json` or `pwd/.wpscan/scan.yml`
+1. `$XDG_CONFIG_HOME/wpscan/scan.json` or `$XDG_CONFIG_HOME/wpscan/scan.yml`
+2. `~/.config/wpscan/scan.json` or `~/.config/wpscan/scan.yml`
```
Initsogar: same as https://github.com/wpscanteam/wpscan/pull/2029#discussion_r3244863297

```suggestion
1. `$XDG_CONFIG_HOME/wpscan/scan.json` or `$XDG_CONFIG_HOME/wpscan/scan.yml` (if `XDG_CONFIG_HOME` is set)
2. `~/.config/wpscan/scan.json` or `~/.config/wpscan/scan.yml` (if `XDG_CONFIG_HOME` is not set)
```

README.md
```diff
@@ -152,6 +152,10 @@ Up to **25** API requests per day are given free of charge, that should be suita
 
 WPScan can load all options (including the `--url`) from configuration files, the following locations are checked (order: first to last):
 
+- `$XDG_CONFIG_HOME/wpscan/scan.json`
+- `$XDG_CONFIG_HOME/wpscan/scan.yml`
+- `~/.config/wpscan/scan.json`
+- `~/.config/wpscan/scan.yml`
```
Initsogar: Just to manage expectations, I think we should clarify in the documentation that only 6 locations are checked per run, not all 8. The first 2 locations will depend on whether XDG_CONFIG_HOME is set.

```suggestion
- `$XDG_CONFIG_HOME/wpscan/scan.json` (if `XDG_CONFIG_HOME` is set)
- `$XDG_CONFIG_HOME/wpscan/scan.yml` (if `XDG_CONFIG_HOME` is set)
- `~/.config/wpscan/scan.json` (if `XDG_CONFIG_HOME` is not set)
- `~/.config/wpscan/scan.yml` (if `XDG_CONFIG_HOME` is not set)
```

Initsogar (approved):
Looks good overall. Tested it and works as advertised!

The implementation is clean and follows the XDG spec correctly. I just left a small suggestion on the docs update.

> If the PR is accepted, please complete https://github.com/wpscanteam/wpscan/blob/master/spec/lib/controllers_spec.rb#L139-L154

Looks like this has already been addressed in https://github.com/wpscanteam/wpscan/pull/2029/commits/2e88fe18077c77598285d88593d71c4924bfd731 👍🏽

PR #2030, opened 2026-05-13
variabilise hardcoded value

noraj:
Related to https://github.com/wpscanteam/wpscan/pull/1815

## Proposed changes

Avoid hardcoding wpscan, use the same syntax as in https://github.com/wpscanteam/wpscan/blob/1845814cf716332b4a70cde9f167d3455fe1e0db/lib/wpscan/controllers.rb#L21

Homogeneity, less hardcoded values

## Testing instructions

Run tests, nothing should change

## Licensing

### DIFF
--- a/lib/wpscan.rb
+++ b/lib/wpscan.rb
@@ -65,8 +65,8 @@ module WPScan
   # XDG support for DB_DIR
   # If the legacy path exists, it will be used. Otherwise, we'll use
   # $XDG_CACHE_HOME/wpscan/db or ~/.cache/wpscan/db when XDG_CACHE_HOME is unset.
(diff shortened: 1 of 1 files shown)


noraj:
Note: I'm not sure which import is missing to avoid the [undefined error](https://github.com/wpscanteam/wpscan/actions/runs/25807085783/job/75812764905?pr=2030)

ArSn (changes requested):
I think the issue is rather that the app_name is defined further down in this same file, could either be switched out place-wise or we just leave it at `.wpscan` as that name is not very likely to change anyhow? 🙂

ArSn (approved):
doesn't hurt 🙂

PR #2038, opened 2026-05-15
Standardize on Pathname objects for path handling

Initsogar:
Follow-up of https://github.com/wpscanteam/wpscan/pull/2034

## Proposed changes

* Remove `.to_s` from 4 enumeration CLI option defaults (timthumbs-list, config-backups-list, db-exports-list, backup-folders-list)
* Remove `.to_s` from 3 DB path getter methods (wp_fingerprints_path, metadata_file, df_file)
* Make `tmp_directory` method return `Pathname` instead of `String`
* Convert `File.join(tmp_directory, ...)` calls to `tmp_directory.join(...)` for cookie-jar and cache-dir defaults
* Update return type documentation from `String` to `Pathname` for affected methods
* Update test specs to match the new Pathname return types

#2034 changed `OptPath#validate` to return `Pathname` objects, but several places still explicitly converted paths to strings using `.to_s` (enumeration defaults, DB getters, tmp_directory method). This created a type mismatch where user-provided paths would be Pathname objects, but default values would be strings.

Removing these conversions completes the standardization, eliminates type inconsistency, and enables cleaner path manipulation using Pathname methods. All Ruby file operations accept Pathname objects, making these conversions unnecessary.

## Testing instructions

Trust the CI, and maybe verify output file creation works:

```bash
bundle exec ruby -Ilib bin/wpscan --help --output /tmp/test.txt
```

## Licensing

### DIFF
--- a/app/controllers/core/cli_options.rb
+++ b/app/controllers/core/cli_options.rb
@@ -122,7 +122,7 @@ module WPScan
                           writable: true,
                           readable: true,
                           create: true,
-                          default: File.join(tmp_directory, 'cookie_jar.txt')),
+                          default: tmp_directory.join('cookie_jar.txt')),
           OptBoolean.new(['--expect-saml',
                           'Expect SAML authentication to be required. ' \
                           'When the target redirects to a SAML IdP, an interactive browser ' \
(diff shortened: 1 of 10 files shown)


Initsogar:
rebasing / fixing conflicts

ArSn (approved):
Much cleaner 👍🏻

PR #2053, opened 2026-05-26
Migrate to API v4

Initsogar:
Related to #

## Proposed changes

* Bump the WPScan API base URL from `/api/v3/` to `/api/v4/` (`lib/wpscan/db/vuln_api.rb`), and update the `from_vuln_api?` request-tracking test to match.
* Rewrite the misleading `# Remove this when api/v4 is up` comment on the `/latest` guard. The guard stays: it prevents a crash when a slug or WordPress version equals the literal string `"latest"` and would collide with the `/{plugins,themes,wordpresses,all}/latest` list endpoints.
* Read the v4 `affected_versions` array internally on `WPScan::Vulnerability` (one entry per backported branch). `WpItem#vulnerable_to?` now matches if any range covers the detected version, which fixes the false negative where a site on an older but still-supported branch was reported as not vulnerable.
* Resolve `fixed_in` / `introduced_in` to the single range that covers the detected version and expose them as scalars, so JSON, CLI, and SARIF output are unchanged from v3. When the version can't be detected (or sits past every fix), fall back to the highest range, mirroring the previous single-value behavior.
* Keep the `Vulnerability` constructor accepting scalar `fixed_in:` / `introduced_in:` (normalized into a single range internally), so non-API callers such as `Timthumb` stay unchanged.
* Update specs and fixtures: input fixtures adopt the v4 `affected_versions` shape; output fixtures keep the scalar `fixed_in`; add multi-range coverage in `plugin_spec.rb` and `vulnerability_spec.rb` (branch selection and the undetected-version fallback).

v4 API migration

## Testing instructions

* Trust the CI, and:
* Manual scan against a site running a plugin with a multi-range backported vulnerability, where the installed version is on an older still-supported branch (`ruby -Ilib bin/wpscan --url https://target.example/ --api-token <token> --enumerate vp`). The plugin should be reported as vulnerable, and the vulnerability should show the `Fixed in:` value for the branch the installed version is on (one line, not every branch). Confirm `--format json` still emits a scalar `fixed_in`, and `--format sarif` still shows `fixed in X` in the rule message.

## Licensing

### DIFF
--- a/app/models/wp_version.rb
+++ b/app/models/wp_version.rb
@@ -58,7 +58,9 @@ module WPScan
         @vulnerabilities = []
 
         Array(db_data['vulnerabilities']).each do |json_vuln|
-          @vulnerabilities << Vulnerability.load_from_json(json_vuln)
+          vulnerability = Vulnerability.load_from_json(json_vuln)
+          vulnerability.detected_version = self
+          @vulnerabilities << vulnerability
         end
 
         @vulnerabilities
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,10 +20,10 @@ jobs:
 
     steps:
     - name: Checkout code
-      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+      uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
 
     - name: Set up Ruby ${{ matrix.ruby }}
-      uses: ruby/setup-ruby@afeafc3d1ab54a631816aba4c914a0081c12ff2f # v1
+      uses: ruby/setup-ruby@12fd324f1d0b43274fdc8130f6980590a667c455 # v1
       with:
         ruby-version: ${{ matrix.ruby }}
         bundler-cache: true
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -10,7 +10,7 @@ jobs:
 
     steps:
       - name: checkout sources
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
 
       - name: Set tag to latest
         if: (github.event_name == 'push' && github.ref == 'refs/heads/master') || github.event_name == 'schedule'
@@ -31,7 +31,7 @@ jobs:
           exit 1
 
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@ce360397dd3f832beb865e1373c09c0e9f86d70a # v4
+        uses: docker/setup-qemu-action@06116385d9baf250c9f4dcb4858b16962ea869c3 # v4
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@d7f5e7f509e45cec5c76c4d5afdd7de93d0b3df5 # v4
--- a/.github/workflows/gempush.yml
+++ b/.github/workflows/gempush.yml
@@ -10,9 +10,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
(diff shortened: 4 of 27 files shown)


ArSn (changes requested):
I am not sure if I agree with the approach taken here so I'll request changes for now.

My "beef" with this approach is that we break the contract / make a backward incompatible change. Of course this is only a big issue for users that actually consume the JSON (or similar machine readable) output formats, but still, we're effectively changing the format here.

I also don't think that if you are scanning a specific site, that you are all that interested that a bunch of different versions of the same plugin (but not the version installed) is vulnerable, would you?

So I propose that we change the approach to:

1. Use the v4 API behind the scenes so we get all the version ranges
2. Only if a version matches, we use that as fixed_in / introduced_in versions for the output, because this is the version that we care about for this scan

This would enable us to do both:
- Make use of the v4 API to get all the vulnerable ranges
- Still keep the change backwards compatible in terms of output, so it could be a minor version change

The big question is what we do if the version could not be detected, but maybe then we just fall back to showing the highest version range as before?

Initsogar:
@ArSn thank you for this valuable feedback. Good call, and I agree with the contract concern.

The actual bug we're fixing is the false negative case where a site on an older, still-supported branch is incorrectly marked as not vulnerable. That can be fixed purely by reading all ranges from v4 and selecting the one that covers the installed version. It doesn't require changing the output format.

I'll rework it to:
1. Keep using v4 behind the scenes to get every range.
2. Select the range covering the detected version and populate the existing `fixed_in` / `introduced_in` exactly as before so JSON/SARIF stay backward compatible, and this can ship as a minor version.
3. When the version isn't detected, fall back to the highest range (closest to current v3 behavior).

Initsogar:
@ArSn just pushed the rework based on your feedback. Could you take another look?

Summary of what changed:
- We still call v4 behind the scenes and read every backported range.
- `fixed_in` / `introduced_in` are resolved to the single range that covers the detected version, and surfaced as scalars exactly as before. JSON, CLI, and SARIF output are unchanged from v3, so nothing breaks for machine-readable consumers and this stays a minor.
- When the version can't be detected, we fall back to the highest range, matching the previous behavior.
- Net effect: the underlying false negative is fixed (a site on an older still-supported branch is now matched against the right range) without changing the output contract.

Thanks, again :)

app/models/wp_version.rb
```diff
@@ -58,7 +58,9 @@ def vulnerabilities
         @vulnerabilities = []
 
         Array(db_data['vulnerabilities']).each do |json_vuln|
-          @vulnerabilities << Vulnerability.load_from_json(json_vuln)
+          vulnerability = Vulnerability.load_from_json(json_vuln)
+          vulnerability.detected_version = self
+          @vulnerabilities << vulnerability
```
ArSn, line 63: with this change, I think we need to change [this line](https://github.com/wpscanteam/wpscan/blob/c7ce265748fb14ec8999acc0e111614c73df1814/app/models/version.rb#L37) to:

```ruby
        other = Version.new(other) unless other.is_a?(Version) # handle potential '.1' version
```

because otherwise if we have a `WpVersion` object that [checks if the version is valid in the constructor](https://github.com/wpscanteam/wpscan/blob/4846f68c97ade4b4adbac20eb476ee05e64f78cb/app/models/wp_version.rb#L10) and to do that [checks the local database](https://github.com/wpscanteam/wpscan/blob/4846f68c97ade4b4adbac20eb476ee05e64f78cb/app/models/wp_version.rb#L28-L34), so if you have an old database and then a WpVersion object gets compared in `<=>` then it throws that kind of error.

Ideally we also add a test to cover that case, to be on the safe side 🙂  (and it is the easiest way to trigger the error conditions)

ArSn (changes requested):

app/models/wp_version.rb
Initsogar: Woah, nice catch! Fixed it and added a test to cover it :) -> https://github.com/wpscanteam/wpscan/pull/2053/commits/a094696c8c90110dc77912fcc4a559d138c8c480

ArSn (approved):
Thanks 🙂

PR #2058, opened 2026-06-12
Changed docker release workflow so that we also always push the "latest" tag along with the versioned tag for the new version.

ArSn:
## Proposed changes

  * Update the Docker release workflow so a published release pushes both the versioned tag (e.g. `4.0.0`) and the `latest` tag in the same build.
  * Replace the single `DOCKER_TAG` env var with a multi-line `DOCKER_TAGS` var holding the full `wpscanteam/wpscan:<tag>` references, using a heredoc for the release case.
  * Update the guard step and `docker/build-push-action` `tags` input to consume the new `DOCKER_TAGS` variable.

  * Previously a release only published the versioned tag, leaving `latest` stale until we fixed that manually.
  * Tagging both in a single multi-arch build keeps `latest` in sync with the released version and avoids a second build.

  ## Testing instructions

I guess we'll test it live with the next release? 🙂

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -15,19 +15,24 @@ jobs:
       - name: Set tag to latest
         if: (github.event_name == 'push' && github.ref == 'refs/heads/master') || github.event_name == 'schedule'
         run: |
-          echo "DOCKER_TAG=latest" >> $GITHUB_ENV
+          echo "DOCKER_TAGS=wpscanteam/wpscan:latest" >> "$GITHUB_ENV"
 
-      - name: Set tag to release name
+      - name: Set tags to release name and latest
         if: github.event_name == 'release' && startsWith(github.ref, 'refs/tags/')
         env:
           RELEASE_TAG_NAME: ${{ github.event.release.tag_name }}
         run: |
-          echo "DOCKER_TAG=$RELEASE_TAG_NAME" >> "$GITHUB_ENV"
+          {
+            echo "DOCKER_TAGS<<EOF"
+            echo "wpscanteam/wpscan:$RELEASE_TAG_NAME"
+            echo "wpscanteam/wpscan:latest"
+            echo "EOF"
+          } >> "$GITHUB_ENV"
 
-      - name: Check if DOCKER_TAG is set
-        if: env.DOCKER_TAG == ''
+      - name: Check if DOCKER_TAGS is set
+        if: env.DOCKER_TAGS == ''
         run: |
-          echo DOCKER_TAG is not set!
+          echo DOCKER_TAGS is not set!
           exit 1
 
       - name: Set up QEMU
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -12,23 +12,23 @@ jobs:
       - name: checkout sources
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
 
-      - name: Set tag to latest
-        if: (github.event_name == 'push' && github.ref == 'refs/heads/master') || github.event_name == 'schedule'
```
ArSn, line 16: this condition never happens anymore because we changed the workflow triggers some time ago

.github/workflows/docker.yml
```diff
@@ -12,23 +12,23 @@ jobs:
...
+      - name: Check release tag is set
+        if: github.event.release.tag_name == ''
         run: |
-          echo "DOCKER_TAG=latest" >> $GITHUB_ENV
+          echo "release tag_name is empty!"
+          exit 1
 
-      - name: Set tag to release name
-        if: github.event_name == 'release' && startsWith(github.ref, 'refs/tags/')
```
ArSn, line 21: and this condition always happens now, so no need to check for it really.

.github/workflows/docker.yml
```diff
@@ -12,23 +12,23 @@ jobs:
...
           RELEASE_TAG_NAME: ${{ github.event.release.tag_name }}
         run: |
-          echo "DOCKER_TAG=$RELEASE_TAG_NAME" >> "$GITHUB_ENV"
-
-      - name: Check if DOCKER_TAG is set
-        if: env.DOCKER_TAG == ''
-        run: |
-          echo DOCKER_TAG is not set!
-          exit 1
+          DELIMITER="EOF_$(openssl rand -hex 16)"
```
ArSn, line 26: a bit overkill probably lol, but why not

.github/workflows/docker.yml
```diff
@@ -12,23 +12,23 @@ jobs:
...
+          {
+            echo "DOCKER_TAGS<<$DELIMITER"
+            echo "wpscanteam/wpscan:$RELEASE_TAG_NAME"
+            echo "wpscanteam/wpscan:latest"
```
Initsogar: I don't expect us to create GH pre-releases (looks like we never did), but the feature is there, and if we ever do it, I understand the workflow would also push the Docker latest tag. Should we guard against that just to be safe?

```suggestion
            if [ "$IS_PRERELEASE" = "false" ]; then
              echo "wpscanteam/wpscan:latest"
            fi
```

(also needs ` IS_PRERELEASE: ${{ github.event.release.prerelease }}` added under `env:` (a few lines up)

Any thoughts?

ArSn: That makes perfect sense, good catch! Done in https://github.com/wpscanteam/wpscan/pull/2058/changes/c54362435cecbe04576bbaadc213ba572d3ae36a

Initsogar (approved):
LGTM!

PR #2073, opened 2026-07-10
Add opt-in local vulnerability database support (--enterprise-db-token)

ArSn:
Related to https://linear.app/a8c/issue/WPSCAN-101

## Proposed changes

Adds an opt-in `--enterprise-db-token TOKEN` option (also settable via the `WPSCAN_ENTERPRISE_DB_TOKEN` environment variable or the `enterprise_db_token` config-file key) that lets enterprise customers scan using a local vulnerability database dump instead of the online API.

When the option is set:

- During the normal database update, WPScan additionally downloads `plugins.json.gz`, `themes.json.gz` and `wordpresses.json.gz` from `enterprise-data.wpscan.org`, authenticated with the token via the `X-DB-JSON-AUTH` header. As with the rest of the database, these are only re-downloaded when their checksums change.
- Plugin, theme and WordPress vulnerability data is then read from those local dumps instead of querying the API — no per-request API calls.
- The API status is reported as plan `enterprise` with `Unlimited` requests remaining.

`--enterprise-db-token` and `--api-token` are mutually exclusive (including via their environment variables); supplying both raises an error. The token is redacted from the logged command line, and `vp`/`vt` (vulnerable plugin/theme) enumeration works without an API token when this mode is active. When no enterprise token is supplied, behavior is unchanged.

## Why

Enterprise customers who mirror the vulnerability database locally can run scans without letting us know which extensions they are detecting on their scanned target sites; which is a privacy concern for some customers. This brings that capability into the official CLI as a supported, opt-in feature.

## Testing instructions

* Install [mitmproxy](https://formulae.brew.sh/cask/mitmproxy)
* Run mitmdump: `mitmdump -p 8081 --ssl-insecure`
* Scan a site with some findings on it using an API token, be sure to specify `--proxy http://localhost:8081 --disable-tls-checks` for visibility -> Here you can see how the requests for plugin/theme vulnerability checking normally go to the WPScan API
* Now scan the site again with the same parameters but use the new `--enterprise-db-token` (and not the API token) -> observe how no vulnerability check requests go to the WPScan API anymore, only requests to download the database dump
* Verify the things it finds are the same (barring small variances due to the DB dump being updated "only" daily)

🤖 Generated with [Claude Code](https://claude.com/claude-code) and heavily edited


app/controllers/vuln_api.rb
```diff
@@ -22,9 +31,13 @@ def cli_options
       end
 
       def before_scan
-        return unless ParsedCli.api_token || ENV.key?(ENV_KEY)
+        raise Error::ConflictingApiTokens if enterprise_db_token && api_token
```
Initsogar: The mutual-exclusion check lives in `VulnApi#before_scan`, which runs after core DB update, so both tokens still trigger a full DB + enterprise-dump download before the scan aborts. And on the `--update`-only path (no `--url`) it exits before `VulnApi` runs, so the conflict is never reported at all. Is that intended, or worth validating before the DB update so it fails fast? Not blocking either way.

Observed on a scan with both tokens (valid enterprise token, so the dumps actually downloaded):

```
[i] Updating the Database ...
[i] Update completed.

Scan Aborted: --api-token and --enterprise-db-token are mutually exclusive, please provide only one (this also applies to the WPSCAN_API_TOKEN and WPSCAN_ENTERPRISE_DB_TOKEN environment variables).
```

The `--update`-only path (no URL) downloads the same DB and exits without ever raising `ConflictingApiTokens`.

Initsogar (approved):
LGTM. Just flagged a [non-blocker](https://github.com/wpscanteam/wpscan/pull/2073/files#r3624359576) I found that might be intended :)

app/controllers/vuln_api.rb
ArSn: oh oops, well I guess that should be checked before updating, but as you said it does not really hurt either way.

I'll still change it to make more sense so that the validation happens first 🙂

ArSn: done in https://github.com/wpscanteam/wpscan/pull/2073/changes/7d986851dfe85ce9b2818162d15963ba9f79f096

Initsogar (approved):
Looks good! Thanks for addressing the suggestion :)

Issue #2079, opened 2026-08-13
API v4 plugin/theme lookups return 401 despite valid token (v3 works, v4 undocumented)


0xAnk:
### Subject of the issue
API v4 plugin/theme vulnerability lookups return 401 Unauthorized despite a valid, working API token (v3 works fine with the same token).

### Your environment
Version of WPScan: 4.1.0
Version of Ruby: ruby 3.3.8 (2025-04-09 revision b200bad6cd) [x86_64-linux-gnu]
Operating System (OS): Kali Linux 2026.3

### Steps to reproduce
1. Obtain a free-tier API token from wpscan.com/profile
2. Run:
   `wpscan --url http://blog.inlanefreight.local --enumerate p --plugins-list mail-masta --api-token <token> --clear-cache -v`
3. Target has Mail Masta 1.0 installed, which has two known, indexed vulnerabilities in the WPScan database ([CVE-2016-10956](https://wpscan.com/vulnerability/5136d5cf-43c7-4d09-bf14-75ff8b77bb44/) LFI, [CVE-2017-6095](https://wpscan.com/vulnerability/c992d921-4f5a-403a-9482-3131c69e383a/) SQLi)
4. Proxy the same command through Burp Suite with --proxy http://127.0.0.1:8080 --disable-tls-checks to inspect the raw API traffic

Note: target is on a HackTheBox lab VPN. Connectivity to the target itself works fine; the issue reproduces identically via direct `curl` to wpscan.com's API (bypassing wpscan entirely), both on and off the VPN, ruling out VPN/network interference as the cause.

### Expected behavior
The scan should display a [!] X vulnerabilities identified block under the mail-masta plugin entry, listing the known CVEs, since this data is confirmed present in the WPScan database.

### Actual behavior
No vulnerability data is displayed. The scan prints [+] WPScan DB API OK and reports a plausible request count / quota, but no vulnerabilities are shown for any plugin, even ones with confirmed, indexed CVEs.

Proxying through Burp shows the actual cause: WPScan CLI calls GET https://wpscan.com/api/v4/plugins/mail-masta, which returns 401 Unauthorized (three consecutive attempts) — despite the same token succeeding on:

GET https://wpscan.com/api/v4/status?version=4.1.0 → 200 OK
GET https://wpscan.com/api/v3/plugins/mail-masta → 200 OK, with full correct vulnerability data (verified via direct curl)

```
[...]

PR #2081, opened 2026-08-19
Run the CLI integration tests with a free API token too

ArSn:
Related to #2079

## Proposed changes

* Add a second integration test scan of the same demo site using a free plan API token, from the new `WPSCAN_FREE_API_TOKEN` secret.
* Assert the API plan each run actually resolved to, via a new `EXPECTED_PLAN` env var checked both before the scan and against `vuln_api.plan` in the JSON output.

* CI only ever scanned with the enterprise token, so regressions specific to the free plan were invisible here. Free plan tokens lost access to the v4 plugin/theme endpoints, and it took a user report (#2079) to surface it.
* The plan assertion keeps the two runs from silently collapsing into one: without it, a wrong value in either secret would just run the same plan twice and quietly drop the free plan coverage.

## Testing instructions

* CI covers this: the `Run WPScan against test site with a free API token and verify results` step should pass on all three Ruby versions and report `✓ Plan: free`, while the enterprise step reports `✓ Plan: enterprise`.
* Check the uploaded `integration-test-results-ruby-*` artifacts contain both `scan-results.json` and `free-scan-results.json`, with a `vuln_api.plan` of `enterprise` and `free` respectively.

🤖 Generated with [Claude Code](https://claude.com/claude-code) and heavily edited


Initsogar (approved):
Nice one. Thanks for also adding a preflight check 🙌

